PR #33, opened 2019-12-02
print warning when slack client is disabled instead of failing

rawxfish:


hunts/views.py
```diff
@@ -93,17 +93,21 @@ def post(self, request, pk):
             # TODO(asdfryan): Add error handling and refactor into slack lib.
             slack_client = SlackClient.getInstance()
             channel_id = slack_client.create_or_join_channel(name)
+            if channel_id is None:
+                messages.warning(request, "Slack channel not created")
```
erwa: Oh cool, `messages` is some nifty Django functionality.

smallboard/templates/base.html
```diff
@@ -53,7 +53,12 @@
 </nav>
 
 {% for message in messages %}
-<div class="alert alert-danger" style="white-space: pre-line" role="alert">{{ message }}</div>
+{% if message.level == DEFAULT_MESSAGE_LEVELS.ERROR %}
+    <div class="alert alert-danger" style="white-space: pre-line" role="alert">{{ message }}</div>
+{% endif %}
+{% if message.level == DEFAULT_MESSAGE_LEVELS.WARNING %}
+    <div class="alert alert-warning" style="white-space: pre-line" role="alert">{{ message }}</div>
```
erwa: Can you paste a screenshot of what this looks like?

erwa:
I'm going to sleep now, but feel free to merge!

hunts/views.py
```diff
@@ -93,17 +93,21 @@ def post(self, request, pk):
...
                     url=puzzle_url,
                     hunt=hunt,
                     sheet=sheet,
-                    channel=channel_id
+                    channel=channel_id if channel_id else name
                 )
                 # Announce new puzzle is available on slack.
-                slack_client.announce_puzzle_creation(name, channel_id, is_meta)
+                if channel_id:
+                    slack_client.announce_puzzle_creation(name, channel_id, is_meta)
```
asdfryan: Hmm ideally slack_client.announce_puzzle_creation() should just do nothing when slack API token is missing. I think I'd rather make that fix instead.

smallboard/templates/base.html
rawxfish: Screen Shot 2019-12-01 at 11 57 08 PM

rawxfish: Screen Shot 2019-12-02 at 12 00 03 AM

Issue #32, opened 2019-12-02
Ability to add tags to puzzles


erwa:
Things like "backsolved", "cryptic", "interaction puzzle", etc. This information could also be put in a "Notes" section per puzzle but tags can be useful for categorizing puzzles (we could allow sorting by tag(s) or only showing puzzles with certain tag(s)).

I think free-form tags would be easiest for users, though there might be variations in spelling and terminology that make filtering harder. But given there's not THAT many puzzles, I think people can figure it out ad-hoc.

For an example of tagging functionality that other systems have used, see this video of Codex Blackboard: https://www.youtube.com/watch?v=oHMN3GujuzQ

PR #35, opened 2019-12-02
Add Puzzle tagging functionality

akirabaruah:
Lets users tag puzzle objects with arbitrary strings for easier filtering. Requires a database migration to include the new tag models included by the Django taggit plugin.

BUG: #32


puzzles/templates/puzzles_table.html
```diff
@@ -8,6 +8,7 @@
             <th scope="col">Status</th>
             <th scope="col">Sheet</th>
             <th scope="col">Slack</th>
+	    <th scope="col">Tags</th>
```
erwa: \*GASP\* Indentation is off. (Not a big deal, I'm just nitpicking.)

PR #37, opened 2019-12-02
Display first-name last-name rather than username when available

erwa:
Username only:
username-only

First-name + last-name:
first-name-last-name

### DIFF
--- a/smallboard/templates/base.html
+++ b/smallboard/templates/base.html
@@ -27,7 +27,12 @@
                 <a class="nav-link" href="{% url 'logout' %}">Logout</a>
             </li>
             <li class="nav-item">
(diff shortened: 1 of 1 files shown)


smallboard/templates/base.html
```diff
@@ -27,7 +27,12 @@
                 <a class="nav-link" href="{% url 'logout' %}">Logout</a>
             </li>
             <li class="nav-item">
-                <div class="nav-link disabled">Logged in as {{ user.username }}
+                <div class="nav-link disabled">Logged in as
+            {% if user.first_name %}
```
asdfryan: Nit: Should we check for both first_name and last_name?

erwa: The last_name just won't show if it's empty. No harm done I think.

PR #38, opened 2019-12-03
Enable column-search for a few columns in the puzzles table.

asdfryan:
Tested locally -- see screenshots below.

Original page:
screenshot1

Basic column filter works:
screenshot2

It filters on the right column:
screenshot3

Combinations of filters also work:
screenshot4

### DIFF
--- a/puzzles/templates/puzzles_table.html
+++ b/puzzles/templates/puzzles_table.html
@@ -8,7 +8,7 @@
             <th scope="col">Status</th>
             <th scope="col">Sheet</th>
             <th scope="col">Slack</th>
-	    <th scope="col">Tags</th>
+            <th scope="col">Tags</th>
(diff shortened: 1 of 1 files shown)


erwa:
Niiice!! Looks great.

One thing (only somewhat related): do you have a preference of whether the first column is titled "Puzzle" or "Name". I believe @rawxfish had called it "Name" originally. I was fine with that and the only reason I changed it to "Puzzle" was because I was testing Heroku deployment and wanted to make some small change that was very visible in the UI. I think I like the heading "Name" better (I think puzzle is implicit about everything on Small Board) and wanted to ask what other people thought.

asdfryan:
I think either "Name" or "Puzzle" is fine -- we can change it back.

PR #41, opened 2019-12-04
Restrict puzzle status update

asdfryan:
Restrict UI to only allow changing to SOLVING, STUCK, or EXTRACTION. PENDING/SOLVED should be handled by the answer queue.

### DIFF
--- a/puzzles/forms.py
+++ b/puzzles/forms.py
@@ -24,7 +24,7 @@ class StatusForm(forms.ModelForm):
         fields = ["status"]
 
     status = forms.ChoiceField(
-        choices=[(status, status) for status in Puzzle.STATUS_CHOICES],
+        choices=[(status, status) for status in Puzzle.ALL_STATUSES],
         widget=forms.Select(attrs={"onChange":'this.form.submit();', 'class': 'form-control form-control-sm'}))
 
 
--- a/puzzles/models.py
+++ b/puzzles/models.py
@@ -16,7 +16,7 @@ class Puzzle(models.Model):
     STUCK = 'STUCK'
     EXTRACTION = 'EXTRACTION'
     ALL_STATUSES = [SOLVING, PENDING, SOLVED, STUCK, EXTRACTION]
-    # Users should only be able to change status to one of these 2.
+    # Users should only be able to change status to one of these 3.
     VISIBLE_STATUS_CHOICES = [SOLVING, STUCK, EXTRACTION]
 
     status = models.CharField(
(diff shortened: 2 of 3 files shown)


puzzles/forms.py
```diff
@@ -24,7 +24,7 @@ class Meta:
         fields = ["status"]
 
     status = forms.ChoiceField(
-        choices=[(status, status) for status in Puzzle.STATUS_CHOICES],
+        choices=[(status, status) for status in Puzzle.ALL_STATUSES],
```
rawxfish: should this just be VISIBLE_STATUS_CHOICES?

puzzles/models.py
```diff
@@ -14,11 +14,14 @@ class Puzzle(models.Model):
     PENDING = 'PENDING'
     SOLVED = 'SOLVED'
     STUCK = 'STUCK'
-    STATUS_CHOICES = [SOLVING, PENDING, SOLVED, STUCK]
+    EXTRACTION = 'EXTRACTION'
+    ALL_STATUSES = [SOLVING, PENDING, SOLVED, STUCK, EXTRACTION]
+    # Users should only be able to change status to one of these 2.
+    VISIBLE_STATUS_CHOICES = [SOLVING, STUCK, EXTRACTION]
```
erwa: "one of these _3_"

I was debating whether we should differentiate STUCK vs. EXTRACTION but I don't feel strongly either way so fine with having separate options for them.

puzzles/forms.py
erwa: Also curious: how come "status" gets repeated twice: (status, status)? What is that for?

asdfryan: @erwa -- https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices

The first is the actual model field that gets set, while the second is the human-readable string shown in the UI. In our case, the two are the same.

@rawxfish -- Actually no. It's a bit counter-intuitive, but these is the essentially the default choices. In puzzle_extras.py, we restrict choices to the VISIBLE_STATUS_CHOICES if the right conditions are met. There are other ways to do this, but I felt this was the easiest.

puzzles/models.py
asdfryan: Addressed.

puzzles/forms.py
erwa: Thanks for the explanation.

Ah, so here we are defining the "ModelForm" so we need to list all possible values for status, whereas puzzle_extras.py is for rendering the template that gets shown to the user, so there we want to filter the choices.

PR #43, opened 2019-12-05
various small answer tweaks

rawxfish:
- warning for duplicate answers
- capitalize and remove whitespace from guesses
- attempt to handle multiple guesses submitted in succession in a reasonable manner
- table styling


puzzles/templatetags/puzzle_extras.py
```diff
@@ -35,6 +35,8 @@ def get_title(puzzle):
 def puzzle_class(puzzle):
     if puzzle.status == Puzzle.PENDING:
         return "table-warning"
+    elif puzzle.status == Puzzle.EXTRACTION:
+        return "table-danger"
```
erwa: What do you think about using yellow (table-warning) for EXTRACTION and STUCK and something like blue for PENDING?

puzzles/models.py
```diff
@@ -35,9 +37,17 @@ def set_answer(self, answer):
...
 
-    def clear_answer(self):
+    def clear_answer(self, guess):
+        # puzzle already solved by different guess, so this current guess does not affect state
+        if self.status == Puzzle.SOLVED and self.answer != guess:
+            return
+
         self.answer = ''
-        self.status = Puzzle.SOLVING
+        if self.guesses.filter(Q(status=Answer.NEW) | Q(status=Answer.SUBMITTED)):
```
asdfryan: Where is guesses defined?

erwa: I was wondering about this, too, and apparently it's a reverse mapping defined by the `related_name` field in `answers/model.py`:

```
class Answer(models.Model):
    ...
    puzzle = models.ForeignKey('puzzles.Puzzle', on_delete=models.CASCADE, related_name="guesses")
```

I think this means that given a `Puzzle`, you can retrieve all `Answer`s that have been submitted for the puzzle by using the `guesses` field.

Ref: https://stackoverflow.com/questions/2642613/what-is-related-name-used-for-in-django

asdfryan: Ah thanks for explaining.

puzzles/templatetags/puzzle_extras.py
rawxfish: Sure, I dont really have any strong opinions about the styling choices.

answers/views.py
```diff
@@ -41,8 +42,14 @@ def post(self, request, hunk_pk, answer_pk):
...
 
         if status_form.is_valid():
-            answer = get_object_or_404(Answer, pk=answer_pk)
-            answer.set_status(status_form.cleaned_data["status"])
+            guess = get_object_or_404(Answer, pk=answer_pk)
+            status = status_form.cleaned_data["status"]
+            if status == Answer.CORRECT and len(guess.puzzle.answer) > 0:
+                messages.warning(request,
+                    '{} was already marked as solved with the answer "{}"\n'.format(guess.puzzle, guess.puzzle.answer) +
+                    'We won\'t stop ya, but please think twice.')
```
erwa: I think this message gets shown after the form has already been submitted. Is it possible to show this warning (perhaps a pop-up confirmation box) when Submit is pressed and before the POST request is sent? The confirmation box could say something like what you have above plus "Are you sure you still want to submit this answer?"

rawxfish: I'm not really familiar with the javascript, could we go with this for now and add it later?

erwa: Yeah, sure, this is just a nice-to-have, not critical. Fine with this for now.

Issue #30, opened 2019-12-02
Ability to modify puzzle details after creation


asdfryan:
People make typos.

We should support editing of fields post-creation -- most importantly the puzzle name and URL fields which are user-generated. It'd be nice to be able to change slack channel as well if there was a typo in the puzzle name.

asdfryan:
Maybe this feature should only be available to admins so that we don't have to deal with lots of people deleting stuff at the same time.

asdfryan:
Oh one thing that is super important is editing the "meta" status of a puzzle (that is... whether or not it's a meta puzzle). Since it's such a small checkbox, it's easy to forget. Also, sometimes you won't know right off the bat if an existing puzzle is actually a meta puzzle or not.

I can see how this can be tricky from the Django model perspective since a Metapuzzle is actually a subclass of Puzzle. You'd probably have to create a new Metapuzzle, port over the existing puzzle's fields, and then delete the old Puzzle.

asdfryan:
Actually, the more I think about it, the more I'm leaning towards just removing the whole Metapuzzle subclass altogether, and have "is_meta" be a field. The reasons being what I said above, and the fact that it's awkward to search through both Puzzles and Metapuzzles. For example, when we check for channel_id uniqueness, we search through both classes.

@rawxfish -- what do you think?

rawxfish:
Those sound like compelling reasons, let's do it.

PR #45, opened 2019-12-08
Remove MetaPuzzle class in favor of is_meta field in Puzzles

asdfryan:
For rationale, see discussion in https://github.com/cardinalitypuzzles/smallboard/issues/30.


asdfryan:
Note that this also includes changes from https://github.com/cardinalitypuzzles/smallboard/pull/44

PR #48, opened 2019-12-08
Enable filtering by meta

asdfryan:
Example: 
image

image

image

image

image

image


erwa:
I guess this isn't in scope for this PR, but it would be nice if there were some indication of what meta(s) each puzzle is assigned to without having to click "assign meta" to find out. Perhaps we need a "Meta" column? Or maybe once a puzzle has been assigned to meta(s), we replace "assign meta" with a list of the meta(s) that the puzzle has been assigned to. And then we could have a little pencil edit icon next to it in case we needed to update the meta(s) for the puzzle.

PR #52, opened 2019-12-09
Add SLACK_BASE_URL environment variable

erwa:
I've deployed the changes to production and verified that this fixes the issue I was having before where if I was already signed in to a different Slack workspace in my browser and then clicked on a Small Board Slack link, it tried to find the channel in the other workspace where it didn't exist, leading to a "There's been a glitch" error page.


README.md
```diff
@@ -123,6 +123,7 @@ GOOGLE_SHEETS_TEMPLATE_FILE_ID=...
 SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET=...
 
 # for Slack integration
+SLACK_BASE_URL=https://my-workspace.slack.com
```
asdfryan: Should we default to the cardinality-puzzles workspace?

erwa: Ok, that seems fine. Question: is cardinality-puzzles invite-only, or can anyone on the Internet join?

README.md
```diff
@@ -123,6 +123,7 @@ GOOGLE_SHEETS_TEMPLATE_FILE_ID=...
...
 SLACK_API_TOKEN=...
```
asdfryan: Ideally there should be validation to ensure that the base url matches the SLACK_API_TOKEN for the app.

erwa: Will add a TODO for that

README.md
asdfryan: Hmm that's a good point. Let's leave this as is, and leave a comment that it should be the same workspace that the Slack app is installed to.

erwa: Oh ok, I can change it back to no default by default.

I was also thinking having no default makes it easier for another team to use Small Board -- they'll see a warning that no SLACK_BASE_URL is configured and then know to configure it.

PR #58, opened 2019-12-15
Sync tags and metas

rawxfish:
For #32. Syncing is done on 
- puzzle edits (pre & post save)
- puzzle deletions (pre delete)
- metas changes (m2m signal)
- tag addition and removal

There might be some weird cases I didn't think about. Let me know if you guys see anything obvious, if not we can just play around with it until something breaks.


puzzles/templates/show_tags.html
```diff
@@ -33,7 +33,10 @@ <h5 class="modal-title" id="addtag-{{ puzzle.pk }}">Add tag to {{puzzle.name}}</
                     </form>
                 {% endfor %}
 
-                <button type="button" class="btn btn-info btn-sm" data-toggle="collapse" data-target="#collapseTagInput" aria-expanded="false" aria-controls="collapseTagInput"> Add New Tag.. </button>
+                <br>
+                <br>
+
+                <button type="button" class="btn btn-info btn-sm btn-block" data-toggle="collapse" data-target="#collapseTagInput" aria-expanded="false" aria-controls="collapseTagInput"> Add New Tag.. </button>
```
erwa: Could we make the button in the modal popup a + sign like it is on the page itself?

puzzles/templatetags/puzzle_extras.py
```diff
@@ -7,7 +7,11 @@
 DEFAULT_TAGS = [
     ('HIGH PRIORITY', PuzzleTag.RED),
     ('LOW PRIORITY', PuzzleTag.YELLOW),
-    ('BACKSOLVED', PuzzleTag.BLUE)
+    ('BACKSOLVED', PuzzleTag.GREEN),
+    ('WORD', PuzzleTag.WHITE),
+    ('LOGIC', PuzzleTag.WHITE),
+    ('TECHNICAL', PuzzleTag.WHITE),
+    ('SLOG', PuzzleTag.GRAY),
```
erwa: Haha

puzzles/models.py
```diff
@@ -41,10 +43,20 @@ class PuzzleTag(TagBase):
...
+        GRAY: 3,
+        BLUE: 4,
+        GREEN: 5,
+        YELLOW: 6
+    }
     color = models.CharField(
         max_length=10,
         choices=COLORS,
         default=BLUE)
+    is_meta = models.BooleanField(default=False)
```
erwa: I suggest adding a comment that this is an internal field not visible or modifiable from the UI.

puzzles/models.py
```diff
@@ -142,6 +157,42 @@ def clear_answer(self, guess):
...
+        PuzzleTag.objects.filter(name=instance.name).filter(is_meta=True).delete()
+
+
+def update_tags_pre_delete(sender, instance, **kwargs):
+    if instance.is_meta:
+        PuzzleTag.objects.filter(name=instance.name).delete()
+
+
+post_save.connect(update_tags_post_save, sender=Puzzle)
+pre_delete.connect(update_tags_pre_delete, sender=Puzzle)
```
erwa: Cool hooks!

puzzles/templates/show_tags.html
erwa: Saw the new button on Sushi's local deployment. Small nit: can we use "..." (three dots) or no dots at all?

rawxfish: Screen Shot 2019-12-15 at 4 07 52 AM
looks like this now

puzzles/models.py
```diff
@@ -142,6 +158,64 @@ def clear_answer(self, guess):
         self.save()
 
 
+# Hooks for syncing metas and tags
+
```
asdfryan: This file is getting rather large. I think we should separate tag specific logic into a separate file.

I think it's fine to do in a separate PR.

puzzles/models.py
```diff
@@ -142,6 +158,64 @@ def clear_answer(self, guess):
...
+def update_tags_pre_save(sender, instance, **kwargs):
```
asdfryan: What is sender? What is instance? 

It seems instance is a puzzle object -- can we rename to "puzzle" for more clarity?

puzzles/views.py
```diff
@@ -128,19 +129,35 @@ def add_tag(request, pk):
                 name=form.cleaned_data["name"],
                 defaults={'color' : form.cleaned_data["color"]}
             )
-            puzzle.tags.add(tag)
+            if tag.is_meta:
+                # the post m2m hook will add tag
+                puzzle.metas.add(Puzzle.objects.get(name=tag.name))
```
asdfryan: What if puzzles.metas already includes this meta? (is this even possible)

puzzles/models.py
rawxfish: sender is the Puzzle class and instance is the instance of the Puzzle that just saved.  I used the variable names from the documentation https://docs.djangoproject.com/en/2.2/ref/signals, but I'll change it to make it more clear.

puzzles/templates/show_tags.html
erwa: Thanks. That looks good!

puzzles/views.py
rawxfish: It's possible that someone adds a tag at the same time that another person adds a puzzle to a meta, but either way one puzzle.metas.add will go through and trigger the m2m hook to sync the tags.

PR #59, opened 2019-12-15
Implement intelligent meta sorting and meta subtree collapsing

asdfryan:
image

Collapsed view:
image


puzzles/tests.py
```diff
@@ -1,3 +1,64 @@
 from django.test import TestCase
 
-# Create your tests here.
+from hunts.models import *
+from .models import *
+from .puzzle_tree import *
+
+class TestPuzzleTree(TestCase):
```
erwa: Our first tests :-O!

I wonder if we can configure these to run automatically for every pull request (e.g.: using CircleCI or TravisCI).

PR #60, opened 2019-12-15
Add meta cycle detection

asdfryan:
We need meta cycle detection because our puzzles table view is a tree. If there is a cycle, it will fail while trying to render the table.

Test adding meta via tag:
image

Test adding meta via assign metas:
image


puzzles/models.py
```diff
@@ -166,6 +166,17 @@ def has_assigned_meta(self):
         return len(self.metas.all()) > 0
 
 
+# Used for cycle detection before adding an edge from potential ancestor to child.
+# We cannot have cycles, otherwise the PuzzleTree sorting will break.
+def is_ancestor(potential_ancestor, child):
+    if child.pk == potential_ancestor.pk:
+        return True
+    if not child.has_assigned_meta(): False
```
erwa: Should this be `return False`?

PR #64, opened 2019-12-15
Support filtering by tags

asdfryan:
image

image

image

image


erwa:
Oh, hmm, Sushi pointed out the search boxes actually scroll, but if your browser window is too small, you still have to scroll down a bit to see them. But this is a minor issue.

PR #73, opened 2019-12-15
Fix meta search

asdfryan:
image

### DIFF
--- a/puzzles/templates/puzzles_table.html
+++ b/puzzles/templates/puzzles_table.html
@@ -100,8 +100,8 @@ $(document).ready(function() {
     // metas).
     $.fn.dataTableExt.ofnSearch['name-input'] = function(value) {
         var search_str = ""
-        search_str += $(value).find("a").text();
-        search_str += $(value).find(".badge").text();
+        search_str += $(value).find(".puzzle-title").text();
+        search_str += $(value).find(".badge-meta").text();
         return search_str;
     };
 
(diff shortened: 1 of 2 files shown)


puzzles/templates/puzzles_table.html
```diff
@@ -180,6 +182,16 @@ <h5 class="modal-title" id="metatitle-{{ p.pk }}">Assign Meta</h5>
         return search_str;
     };
 
+    // Search by the title of the puzzle, and the term "META" if it's a
+    // metapuzzle.
+    $.fn.dataTableExt.ofnSearch['name-input'] = function(value) {
```
erwa: Hmm, don't really understand the comment above relative to the code below. What's "title" and what's "META" in the code?

asdfryan: Datatable implements pure substring search and filtering based on the search token (by default, the raw HTML as string).

Here I am rewriting the search token to be the title of the puzzle, with "meta" appended to it if it's a meta puzzle.

asdfryan: I've updated the comment to be a little more clear.

asdfryan: I've also changed the jquery to be less magical and more sensible (and added the relevant class tags to the html).

erwa: Ah, thank you for that. Was confused what those 'a' and '.badge' references were.

Issue #84, opened 2019-12-16
less modals


rawxfish:
Use one modal for each action, instead of one modal per action per puzzle.

erwa:
Also see https://github.com/cardinalitypuzzles/smallboard/issues/78#issuecomment-565883469.

PR #85, opened 2019-12-16
use only one meta select modal

rawxfish:
For #84. This works, not sure if this is best practices or performant..


erwa:
https://github.com/cardinalitypuzzles/smallboard/pull/86 seems cleaner.

rawxfish:
Getting rid of this in favor of #86

PR #88, opened 2019-12-17
Use JavaScript to fetch tags modal (#84)

erwa:
I followed in the footsteps of giants and copied the approach from #86 to remove pre-generation of all the tags modals and to fetch them on demand from the server.

With this change deployed, Small Board (with its ~70 puzzles) takes about 3 seconds to load (reduced from 5-6 seconds before) and data transferred is 456 KB (down from 1.1 MB):
no-tags-modal

### DIFF
--- a/puzzles/templates/modals/add_tag.html
+++ b/puzzles/templates/modals/add_tag.html
@@ -1,0 +1,22 @@
+<div class="modal fade" id="addtag" tabindex="-1" role="dialog" aria-labelledby="addtag" aria-hidden="true">
+    <div class="modal-dialog" role="document">
+        <div class="modal-content">
+
+        </div>
+    </div>
+</div>
+
+<script type="text/javascript">
+$(function () {
+  $(".addtag").click(function() {
+    var pk = $(this).data("pk");
+    $.get({
+        "url": `/puzzles/add_tags_form/${pk}`,
+        'csrfmiddlewaretoken': '{{ csrf_token }}'
+	}).done(function(data) {
+        console.log(pk)
+        $("#addtag .modal-content").html(data);
+    })
+  })
+});
+</script>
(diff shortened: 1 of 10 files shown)


puzzles/templatetags/puzzle_extras.py
```diff
@@ -114,11 +112,16 @@ def show_tags(puzzle, tag_form, request):
...
         'current_tags': current_tags,
-        'tag_form': tag_form,
+        'tag_form': TagForm(),
         'suggestions': suggestions
     }
     return context
 
+@register.inclusion_tag('show_tags.html')
+def show_tags(puzzle, request):
+    return create_tag_context(puzzle)
```
rawxfish: I think it might be better to split create_tag_context up into the part that the templatetags needs for show_tags.html (current_tags and puzzle) and the part that tags_form.html needs(puzzle, suggestions, and tag_form). It feels weird to me that the views file imports templatetags

puzzles/templates/modals/add_tag.html
```diff
@@ -0,0 +1,22 @@
...
+    </div>
+</div>
+
+<script type="text/javascript">
+$(function () {
+  $(".addtag").click(function() {
+    var pk = $(this).data("pk");
+    $.get({
+        "url": `/puzzles/add_tags_form/${pk}`,
+        'csrfmiddlewaretoken': '{{ csrf_token }}'
```
rawxfish: Is this necessary? I don't think I need that for set_metas, what is different here?

erwa: Oh, guess this isn't necessary. Thanks for catching this. It was just something I added while trying to figure out how to fix a `A {% csrf_token %} was used in a template, but the context did not provide the value.` error

puzzles/templatetags/puzzle_extras.py
erwa: Hmm, alright, I think you're right that we can factor this a little better. I'll create some tag helper methods in utils.py:
* get_tags(puzzle)
* get_all_tags()

and then have templatetags/puzzle_extra.spy and views.py import utils instead

erwa: Refactored into a `tag_utils` module in latest revision.

Issue #89, opened 2019-12-17
Answer Queue page has TemplateSyntaxError "'get_title' did not receive value(s) for the argument(s): 'edit_form'"


erwa:
![answer-queue-error](https://user-images.githubusercontent.com/544734/70971156-4be4d200-206e-11ea-9b3d-74a9d41dfbeb.png)

PR #99, opened 2019-12-20
Fix Answer Queue page (#89)

erwa:
Will optimize to query the modals from the server on-demand rather than pregenerating them in a follow-up PR.


erwa:
Ignore the CircleCI build error for now, since the repo doesn't have any CircleCI config file until #97 is merged.

Issue #90, opened 2019-12-17
Cannot add any of the following tags: HIGH PRIORITY, BACKSOLVED, LOW PRIORITY, or any <meta> tag


erwa:
![tag-invalid-choice](https://user-images.githubusercontent.com/544734/70971225-7767bc80-206e-11ea-8442-763a8157e897.png)

PR #102, opened 2019-12-20
Fix bug in TagForm

asdfryan:
Fixes https://github.com/cardinalitypuzzles/smallboard/issues/90


asdfryan:
For more context, the bug happens because the same TagForm is used even when selecting one of the reserved tags, and the color you end up selecting is not one of the eligible colors.

The fix is to include all of the colors as eligible in the constructor, and only change the form that get presented in the modal.

Issue #72, opened 2019-12-15
Title should be truncated at some point


maximized:
I entered the title "THISPUZZLETITLEISTHEFULLTEXTOFATLASSHRUGGEDWOULDN'TTHATSUCKIFTHEYDIDTHAT". Instead of showing this full title text, let's just show "THISPUZZLETITLEIST...". This forces me to have to scroll left and right to access features.

You can click on the link or we can have a mouseover that shows the full title. Searching should still search against the full title.

maximized:
If I go over 80 characters:

Environment:


Request Method: POST
Request URL: https://smallboard.herokuapp.com/puzzles/edit/32/

Django Version: 2.2.8
Python Version: 3.6.4
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'puzzles',
 'accounts',
 'hunts',
 'answers',
 'social_django',
 'taggit']
Installed Middleware:
('whitenoise.middleware.WhiteNoiseMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'social_django.middleware.SocialAuthExceptionMiddleware')



Traceback:

File "/app/.heroku/python/lib/python3.6/site-packages/django/db/backends/utils.py" in _execute
  84.                 return self.cursor.execute(sql, params)

The above exception (value too long for type character varying(80)
) was the direct cause of the following exception:

File "/app/.heroku/python/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/app/.heroku/python/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/app/.heroku/python/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

[...]

PR #103, opened 2019-12-20
Truncate puzzle name if >20 characters.

asdfryan:
fixes https://github.com/cardinalitypuzzles/smallboard/issues/72


erwa:
I actually don't think the puzzle name is what's causing the table to be so wide, requiring scrolling. As you can see from the picture below 

table

the puzzle name already wraps to the next line automatically. It seems the main thing that's causing the table to be so wide is that the search boxes at the bottom have some default minimum width.

I think it's useful to have the full (or as much of the puzzle) name preserved as possible. I think we should tweak the text boxes and column widths instead. Or, if we still want to truncate titles, we should preserve the full name and only show a truncated version in the UI (but show the full name on mouseover).

hunts/views.py
```diff
@@ -74,6 +74,8 @@ def post(self, request, pk):
 
         if form.is_valid():
             name = form.cleaned_data["name"]
+            if len(name) > 20:
+                name = name[:20]
```
erwa: I think we should make this 80, which is what we need to fix the 

```
Exception Value: value too long for type character varying(80)
```

error @maximized saw. As for the scrolling (table being too wide), I think that's mainly due to the search boxes at the bottom of the table being too wide -- see my comment here: https://github.com/cardinalitypuzzles/smallboard/pull/103#issuecomment-567924232

asdfryan: I think the discrepancy between what you are seeing and what Max saw is due to the fact that I disabled horizontal table scrolling in https://github.com/cardinalitypuzzles/smallboard/pull/82. 

I think you're right in general -- we should only truncate if we absolutely have to. I've changed my code to reflect this.

Issue #67, opened 2019-12-15
Can't create puzzle without URL


maximized:
Situations in which this might be necessary
1. Hidden puzzle
2. Known meta but not yet unlocked
3. Physical puzzle (although there will probably still be a page you could link to)
4. Left Out does something weird where there is a ClueKeeper hunt within Mystery Hunt

asdfryan:
+1 I think this is important.

We should also consider making URL a non-unique field now that we have edit/delete support.

PR #104, opened 2019-12-20
Support blank puzzle urls

asdfryan:
Fixes https://github.com/cardinalitypuzzles/smallboard/issues/67

Tested locally


erwa:
The "add new puzzle" modal now supports blank URLs, but the "edit puzzle" modal still doesn't.
edit-puzzle-modal

I think there's probably one more place you need to update validation.

erwa:
Also, it would be nice not to show the puzzle name text as a clickable hyperlink if there is no URL. Currently, it still shows up as a link:
empty

I think `title.html` needs to be updated to not include the `<a href=...>` if `puzzle.url` is blank.

asdfryan:
Thanks Anthony for noticing those bugs -- I'll fix them.

asdfryan:
Regarding the href -- right now the link simply redirects to the current page view -- I think that is a reasonable default behavior.

erwa:
Redirecting to the current page is fine, but I think not showing a link at all is clearer.

PR #106, opened 2019-12-20
Disable CSRF checks for add puzzle and add tag to allow scripting

erwa:
I hope you guys don't mind :).


erwa:
Yeah, there’s probably a way. I’ll look into it. Thanks for the suggestion.

erwa:
Updated PR so `@csrf_exempt` is only applied when `DEBUG=True`.

Issue #96, opened 2019-12-19
Logo and favicon for Small Board


erwa:
Every proper web app needs a cool logo and favicon.

erwa:
Closing this now we have a Big Chungus favicon. We can create a new issue if we want to add a logo on the navigation bar, too.

PR #109, opened 2019-12-22
add title and favicon

rawxfish:
For #96:

Screen Shot 2019-12-22 at 1 16 08 PM


erwa:
Thanks for the code clean-up, too.

erwa:
I think Small Board should have a sushi logo in honor of the person who came up with the name and built the initial version.

Issue #105, opened 2019-12-20
Search does not search puzzles under collapsed trees


erwa:
Not sure if this is WAI, but I noticed that when a tree is collapsed, e.g.:

![fear](https://user-images.githubusercontent.com/544734/71279507-f850dd80-2326-11ea-88da-95d79166d0ad.png)

then puzzles in the collapsed tree will not show up in search results. For example, under Fear, there is a "Jeopardy!" puzzle, but if you search for it while Fear is collapsed, the puzzle does not show up.

asdfryan:
It's not that the search isn't searching collapsed puzzles -- the search DOES work, it's just the puzzles remain collapsed. This is pretty horrible yeah. And the solution is difficult. 

The fundamental problem is that datatable and treegrid just don't work well together IMO. If we were to redo this from scratch, I'd probably use something else -- but we probably don't have time at this point.

When a treegrid node is collapsed, the DOMs in the subtree get hidden. If we want search to trigger on these hidden nodes, we'd have to un-hide all of the nodes when the search box is non-empty, somehow remember the previous collapse-state of the table, and then re-hide all of the nodes when the search box returns to being empty. Let me see if this is doable with the current framework.

erwa:
Thanks for fixing this. Definitely an important quality-of-life improvement.

PR #111, opened 2019-12-28
Move column search to top and auto-expand collapsed nodes upon search

asdfryan:
image
image


asdfryan:
This fixes https://github.com/cardinalitypuzzles/smallboard/issues/105

PR #117, opened 2019-12-29
answer queue tests

rawxfish:


answers/tests.py
```diff
@@ -1,3 +1,110 @@
...
+        guess2 = Answer.objects.create(puzzle=self._puzzle, text="guess2")
+
+        guess1.set_status(Answer.CORRECT)
+        self._puzzle.refresh_from_db()
+        self.assertEqual(self._puzzle.status, Puzzle.SOLVED)
+        self.assertEqual(self._puzzle.answer, guess1.text)
 
-# Create your tests here.
+        # this should not affect puzzle status if it is already solved
+        guess2.set_status(Answer.INCORRECT)
```
erwa: Good test

manage.py
```diff
@@ -13,12 +13,15 @@ def main():
     os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'smallboard.settings')
 
     if 'test' in sys.argv:
-        dotenv_file = os.path.join(BASE_DIR, ".env.test")
+        test_env_files = [".env.test.user", ".env.test"]
```
erwa: Assuming the file loaded second overrides the settings in the first, I think we want to load `.env.test.user` second? I think of `.env.test` as the defaults, and then users might want to override or add some settings.

rawxfish: It isn't overrides, it just loads the first file from that list and then breaks out of the loop

erwa: Oh, oops, I didn't notice that. I got you now.

Issue #114, opened 2019-12-28
Optimize performance of main puzzle table view


daiweili:
Currently it takes ~5 seconds to load the main hunts page with ~300 puzzles. An actual hunt would have fewer puzzles, but not by more than an order of magnitude.

Most of the time is spent rendering the table (whether it's the server rendering the HTML table, or the client calculating style and layout as it receives the HTML from the server). This suggests that shipping the puzzle contents to the client as a payload (e.g. JSON) and letting the client assemble the table could be advantageous:
- Let datatables create the table, so only the necessary DOM nodes are created (instead, we get over 1k from the server, which are laid out and rendered before being hidden again by datatables when it paginates the table)
- The HTML payload transferred by the server is 1MB which is substantial (100s of ms). The JSON payload would be smaller. Both would benefit from content compression though (at the cost of some server load).

[Example profile captured from Chrome](https://drive.google.com/file/d/1jcmITI7enr93RwR_qlibMAZtnFuHyhC5/view?usp=sharing)
[Audit from Chrome](https://drive.google.com/open?id=1h8Kw83QybvdV0dy4DMgLK7nUeQO8W8pF)
![Screenshot of same profile](https://i.imgur.com/pRIcmrc.png)

daiweili:
Some more (future) ideas, roughly sorted in order of ease:
- Limit initial fetch of puzzles and do subsequent puzzle fetches in the background
- Do puzzle updates asynchronously (use AJAX or likewise)
- Implement live sync. Example idea: [Use GraphQL subscriptions](https://graphql.org/blog/subscriptions-in-graphql-and-relay/). Switch puzzle view to use something based on React, e.g. [react-redux-grid](https://github.com/bencripps/react-redux-grid). It's only mentioned in this performance issue because implementing live sync should reduce the need to refresh the page.

PR #113, opened 2019-12-28
Performance improvements

daiweili:
Server side: prefetch_related on hunts.puzzles.all() to trim number of SQL queries from O(# puzzles) to O(1)
Client side: use modified version of dataTables to extract text straight out of the DOM API rather than extracting text out of the DOM, re-parsing using jQuery, and then extracting the desired text out of jQuery (each re-parse takes ~5ms, extracting out of the DOM API is > 100x faster)

PR #118, opened 2019-12-30
Send array of data to client, render entire table client side. 

erwa:
Main changes:

* removed puzzle table HTML generation from server side. Instead, we just ship an array of data to the client, do all table rendering using JavaScript client-side. Data transferred reduced from 1.1 MB to 130 KB. Load time is now about 2 seconds on Heroku:
transfer

* removed column-specific search boxes
* removed "Enter Guess" column, combined with "Answer" column, added new "Submit Answer" modal:
answer-modal

* made editing Status an on-demand modal:
edit-status

* fixed bugs where modals weren't working when table was redrawn (e.g., on paging or filtering)
* increase default page size from 25 to 50
* reverted jquery datatables changes in https://github.com/cardinalitypuzzles/smallboard/pull/113/commits/bb6c55738556e5fdfe316fe1cc4e0cb34855e8e2 since they broke datatables searching

Screenshot:
new-ui

For #114.


puzzles/templates/modals/edit_status.html
```diff
@@ -0,0 +1,41 @@
...
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5 class="modal-title" id="edittitle">Edit Status of <span id='puzzletitle'></span></h5>
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <form method="post" action="/puzzles/update_status/">
+                {% csrf_token %}
+                <div class="form-group">
```
daiweili: add modal-body to get padding here

puzzles/templates/puzzles_table.html
```diff
@@ -1,150 +1,174 @@
...
-                <th scope="col"></th>
-                <th scope="col"></th>
-                <th scope="col" class="column_searchable">Tags</th>
-                <th scope="col"></th>
-                <th scope="col"></th>
-            </tr>
-        </tfoot>
-    </table>
-</div>
+<table class="table table-sm tree" id="puzzles" style="width:100%">
```
daiweili: Wrap this in `<div class="px-3 pb-3"></div>` to get some extra padding as before. As is, the table has horizontal scroll.

smallboard/static/jquery.dataTables.js
```diff
@@ -4486,7 +4486,7 @@
 		var columns = settings.aoColumns;
```
daiweili: Feel free to just delete this file and load from the CDN instead

puzzles/templates/puzzles_table.html
```diff
@@ -1,150 +1,174 @@
...
+</table>
 
 <script type="text/javascript">
+var slack_base_url = '{{ slack_base_url }}';
+var data = [
+{% for p, tags, metas, active_users, status_form, puzzle_class in rows %}['{{p.pk}}', '{{p.name}}', '{{p.url}}', '{{p.is_meta}}', '{{active_users|join:", "}}', '{{p.answer}}', '{{p.status}}', '{{p.sheet}}', '{{p.channel}}', {{tags|safe}}, '', '{{puzzle_class}}', ],
```
daiweili: No need to change for now, but in a future change, it'd make sense to expose an endpoint to just fetch the JSON data (rather than the page and having the data be delivered as part of the .html in a script tag)

puzzles/templates/puzzles_table.html
```diff
@@ -1,150 +1,174 @@
...
+                    if (data) {
+                        result = `<a href="${slack_base_url}/app_redirect?channel=${data}" target="_blank">Link</a>`
+                    }
+                    return result;
+                }
+            },
+            {
+                'targets': tags,
+                'render': function(data, type, row, meta) {
+                    result = '';
```
daiweili: nit: var result here and in all the other render functions?

puzzles/templates/modals/edit_status.html
erwa: Ah, thanks, the modal looked weird (no edge padding) and I was wondering why

puzzles/templates/puzzles_table.html
erwa: That looks much better. Thanks for noticing this.

smallboard/static/jquery.dataTables.js
erwa: Good idea

puzzles/templates/puzzles_table.html
erwa: Yeah, I was thinking the same thing. Shouldn't be too hard to add such an endpoint, and then we should be able to refresh the table without an entire page reload. And we can set a pollInterval of, say, 5-10 seconds so the table refreshes automatically.

puzzles/templates/puzzles_table.html
erwa: Updated. Good to avoid globals.

erwa:
Fixed the Answer Queue page. Removed the edit and delete puzzle functionality from the page.
answer-queue

For now, I didn't optimize the answer queue page (all the HTML is still generated server-side). Will optimize it in a follow-up PR.

smallboard/static/moment.min.js
```diff
@@ -0,0 +1 @@
...,arguments).parseZone()},c.localeData=ht,c.isDuration=Rt,c.monthsShort=function(e,t){return vn(e,t,"monthsShort")},c.weekdaysMin=function(e,t,n){return pn(e,t,n,"weekdaysMin")},c.defineLocale=lt,c.updateLocale=function(e,t){if(null!=t){var n,s,i=st;null!=(s=ot(e))&&(i=s._config),(n=new P(t=x(i,t))).parentLocale=it[e],it[e]=n,ut(e)}else null!=it[e]&&(null!=it[e].parentLocale?it[e]=it[e].parentLocale:null!=it[e]&&delete it[e]);return it[e]},c.locales=function(){return s(it)},c.weekdaysShort=function(e,t,n){return pn(e,t,n,"weekdaysShort")},c.normalizeUnits=H,c.relativeTimeRounding=function(e){return void 0===e?An:"function"==typeof e&&(An=e,!0)},c.relativeTimeThreshold=function(e,t){return void 0!==jn[e]&&(void 0===t?jn[e]:(jn[e]=t,"s"===e&&(jn.ss=t-1),!0))},c.calendarFormat=function(e,t){var n=e.diff(t,"days",!0);return n<-6?"sameElse":n<-1?"lastWeek":n<0?"lastDay":n<1?"sameDay":n<2?"nextDay":n<7?"nextWeek":"sameElse"},c.prototype=mn,c.HTML5_FMT={DATETIME_LOCAL:"YYYY-MM-DDTHH:mm",DATETIME_LOCAL_SECONDS:"YYYY-MM-DDTHH:mm:ss",DATETIME_LOCAL_MS:"YYYY-MM-DDTHH:mm:ss.SSS",DATE:"YYYY-MM-DD",TIME:"HH:mm",TIME_SECONDS:"HH:mm:ss",TIME_MS:"HH:mm:ss.SSS",WEEK:"GGGG-[W]WW",MONTH:"YYYY-MM"},c});
```
daiweili: Do you need to serve this? Does this CDN version not work: https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.24.0/moment.min.js

puzzles/templates/modals/edit_status.html
```diff
@@ -1,41 +1,43 @@
...
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <form method="post" action="/puzzles/update_status/">
+                {% csrf_token %}
+                <div class="modal-body">
+                    <div class="form-group">
+                        <select name="status" class="form-control form-control-sm">
+                            <option value="SOLVING" id='SOLVING'>SOLVING</option>
```
daiweili: pre-existing, but why are these id's set to these all caps values? Are they needed?

smallboard/static/moment.min.js
erwa: Oh nice, using CDN is probably better. Will switch to this. Thanks.

puzzles/templates/modals/edit_status.html
erwa: I think I added them to make it easier to get the DOM element using jQuery via `$('#<STATUS>')`. But looks like you can get by value using `$('#editstatus option[value="<STATUS>"])`. Removed the ids.

PR #127, opened 2020-01-05
Generate answer queue table client side

erwa:
Here's what the updated Answer Queue page looks like:

answer-queue

Main changes:

* generate the table client-side (I generated ~1000 guesses locally, and it took less than a second to load):
data-transfer


answers/templates/queue.html
```diff
@@ -14,101 +12,120 @@ <h1>{{ hunt_name }} - Answer Queue</h1>
...
+                    if (row[puzzle_is_meta]) {
+                        result += ' <span class="badge badge-dark badge-meta">META</span>';
+                    }
+                    return result;
+                },
+            },
+            {
+                'targets': guess,
+                'render': function(data, type, row, meta) {
+                    return `<p class="text-monospace">${data}</p>`
```
daiweili: Cool! ES6 template literals! I didn't realize we were using ES6; I guess it's pretty well-supported outside of IE11, and we don't care about that.

erwa: Haha, yeah, I learned that from @rawxfish's PR here: https://github.com/cardinalitypuzzles/smallboard/pull/86/files#diff-a36dd12b9d7f908e5012c640b4326239R30

Also working on these changes, I learned about the `let` keyword (instead of `var`) and `for (x ... of y)` loops.

erwa: It does seem like JavaScript is getting a little better

erwa: I'm actually pretty impressed JavaScript now has these template literals that can be used for multi-line strings. Java doesn't even have such multi-line string literal support!

Issue #49, opened 2019-12-08
Add locking for read-then-write operations


asdfryan:
Django on its own doesn't provide any sort of table-locking.

Given that we are soon adding edit functionality (we already support this for some cases), we need to protect ourselves from race conditions somehow for certain operations. These include:
* Updating answer status, and the relevant puzzle status
* Updating assigned metas (including cycle detection) 

There are 2 approaches: pessimistic and optimistic locking
https://medium.com/@hakibenita/how-to-manage-concurrency-in-django-models-b240fed4ee2

I think given our low QPS and use-case, pessimistic locking is probably better.

PR #132, opened 2020-01-07
Locking

rawxfish:
Pessimistic locking for #49

### DIFF
--- a/puzzles/views.py
+++ b/puzzles/views.py
@@ -86,7 +86,7 @@ def slack_guess(request):
     if puzzle.status == Puzzle.SOLVED:
         return HttpResponse("Puzzle is already solved!")
 
-    answer, created = Answer.objects.get_or_create(puzzle=puzzle, text=answer_text)
+    _, created = Answer.objects.get_or_create(puzzle=puzzle, text=answer_text)
     if not created:
         return HttpResponse("The answer " + answer_text + " has already been submitted.")
     puzzle.status = Puzzle.PENDING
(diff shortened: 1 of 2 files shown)


puzzles/views.py
```diff
@@ -190,24 +197,25 @@ def delete_puzzle(request, pk):
...
         return HttpResponseRedirect(request.META.get('HTTP_REFERER', '/'))
-    puzzle = get_object_or_404(Puzzle, pk=pk)
+    puzzle = get_object_or_404(Puzzle.objects.select_for_update(), pk=pk)
     (tag, _) = PuzzleTag.objects.update_or_create(
         name=form.cleaned_data["name"],
         defaults={'color' : form.cleaned_data["color"]}
     )
     if tag.is_meta:
-        metapuzzle = Puzzle.objects.get(name=tag.name)
+        metapuzzle = get_object_or_404(Puzzle, name=tag.name)
```
erwa: Do we need to use `Puzzle.objects.select_for_update()` here as well?

puzzles/views.py
```diff
@@ -79,22 +81,23 @@ def slack_guess(request):
...
     answer_text = __sanitize_guess(slack_message.get('text'))
     channel_id = slack_message.get('channel_id')
-    puzzle = Puzzle.objects.get(channel=channel_id)
+    puzzle = get_object_or_404(Puzzle.objects.select_for_update(), channel=channel_id)
     print("puzzle: " + str(puzzle))
     if puzzle.status == Puzzle.SOLVED:
         return HttpResponse("Puzzle is already solved!")
-    if Answer.objects.filter(puzzle=puzzle).filter(text=answer_text):
+
+    answer, created = Answer.objects.get_or_create(puzzle=puzzle, text=answer_text)
```
erwa: Nit: I think you can use `_, created` instead.

puzzles/views.py
rawxfish: Hmm, I'm not sure. I didn't think we needed it since we aren't updating metapuzzle.

erwa: Could an inconsistency arise if the metapuzzle were changed to a regular puzzle while this transaction was happening? Then on line 218, we would be treating the now-regular puzzle as the meta for this puzzle?

rawxfish: Yeah you're right, a change to metapuzzle could happen from a third puzzle.

PR #136, opened 2020-01-07
Send answer queue messages to separate #answer-queue channel

erwa:
Changes:
* send answer status update messages to separate #answer-queue channel for easier monitoring by phone person
* add Slack messages for NEW and SUBMITTED status updates
* post CORRECT status update message to puzzle channel as well
* move puzzle creation and CORRECT answer messages from #announcements to #puzzle-announcements

Screenshots:

In puzzle channel:
new-messages

In #answer-queue channel:
answer-queue

### DIFF
--- a/answers/views.py
+++ b/answers/views.py
@@ -89,24 +89,25 @@ class AnswerView(LoginRequiredMixin, View):
         puzzle_channel = answer.puzzle.channel
         message = ''
         if status == Answer.NEW:
-            message = ("NEW answer '%s' has been guessed for puzzle '%s'." %
-                      (answer.text, answer.puzzle.name))
+            message = ("NEW answer '%s' has been guessed for puzzle '%s'."
+                       % (answer.text, answer.puzzle.name))
         elif status == Answer.SUBMITTED:
-            message = ("'%s' has been SUBMITTED for puzzle '%s' on the hunt website." %
-                      (answer.text, answer.puzzle.name))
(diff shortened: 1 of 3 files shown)


answers/views.py
```diff
@@ -77,23 +77,29 @@ def get(self, request, hunt_pk):
...
-                                                      (answer.text.upper(), status))
-
+        message = ''
+        if status == Answer.NEW:
+            message = ("NEW answer '%s' has been guessed for puzzle '%s'." %
+                      (answer.text, answer.puzzle.name))
+        elif status == Answer.SUBMITTED:
+            message = ("'%s' has been SUBMITTED for puzzle '%s' on the hunt website." %
+                      (answer.text, answer.puzzle.name))
+        elif status in [Answer.PARTIAL, Answer.INCORRECT, Answer.CORRECT]:
```
asdfryan: else? are there any other statuses?

erwa: No other statuses. Maybe I’ll add an `else “UNEXPECTED STATUS!”` to avoid forgetting to update this method if we ever add a new status

erwa: Added an else case:
```
        else:
            message = ("Unexpected status '%s' for answer '%s' for puzzle '%s'" %
                      (status, answer.text, answer.puzzle.name))
```

asdfryan: Probably should terminate early here.

erwa: Changed to `logger.error()` and added a `return`.

Issue #115, opened 2019-12-28
Implement live sync of puzzle state


daiweili:
It would be nice if the main hunt dashboard updated live as people added puzzles or modified state of existing puzzles. Currently, seeing the updated puzzle state requires a refresh of the page. See [issue/114](https://github.com/cardinalitypuzzles/smallboard/issues/114#issuecomment-569460240) for an example idea of how we could go about implementing this.

PR #137, opened 2020-01-08
Update puzzles table asynchronously (without page reload)

erwa:
Changes:

* Added new `/hunts/<hunt_id>/puzzles` endpoint used by AJAX to fetch JSON array of puzzle data
* Made puzzles table refresh asynchronously every 10 seconds
* Removed puzzles_extra.py and puzzles_table.html (combined with all_puzzles.html)

For #115.


hunts/templates/all_puzzles.html
```diff
@@ -45,5 +46,276 @@ <h5 class="modal-title" id="addtitle">Add Puzzle</h5>
...
+        initEditPuzzle();
+        initDeletePuzzle();
+        initEditStatus();
+        initSubmitAnswer();
+        initAddTag();
+        initSetMeta();
+    }
+
+    function reload() {
+        $.ajax('/hunts/{{hunt_pk}}/puzzles', {
```
daiweili: This is going to make you lose your position in the table, isn't it? :(

Is there any way we could read that out and put you back to where you were?

Also, this is going to be annoying if you're in the middle of doing something and the page refreshes on you. I suggest doing the following:
- Track whether the user is idle (e.g. https://www.brcline.com/blog/detecting-inactivity-in-javascript or https://github.com/jasonmcleod/jquery.idle)
- When idle and the table is different, go ahead with the update
- If not idle, and the table is different, display an alert https://getbootstrap.com/docs/4.4/components/alerts/ with a button to update the table

erwa: Ah, it actually won't make you lose your position because I use `table.draw(false)` which causes the current page position to be kept, even after redraw/reorder/research: https://datatables.net/reference/api/draw()

erwa: I added a comment and switched to the equivalent `table.draw('full-hold')` for clarity:
```
                    // keep current page position after redraw
                    table.draw('full-hold');
```

erwa: Also, the table is only redrawn when there are new puzzles added (`if (newdata.length !== olddata.length) { ....`), so most of the time, the refresh every 10 seconds won't do anything.

erwa: When only data in a single row changes (e.g.: puzzle name or URL updated, puzzle status changes), only that row is invalidated and redrawn, and this shouldn't affect any search or paging the user has done.

erwa:
Thanks for all your reviews, @daiweili! Much appreciated.

hunts/templates/all_puzzles.html
daiweili: Awesome, thanks for thinking through the cases!

Issue #74, opened 2019-12-15
Populate meta spreadsheet with puzzle answers and links automatically


erwa:
Whenever any puzzles associated with a meta are edited (example: an answer is added), we should automatically add the info to the meta spreadsheet. To avoid conflict with humans editing the spreadsheet, we can add this info to an auto-generated tab that says AUTOGENERATED or DO_NOT_EDIT.

PR #140, opened 2020-01-08
Prepopulate sheet with puzzle and Slack links and meta sheet with feeder puzzles

erwa:
When puzzle is created, the puzzle link and Slack channel link are now pasted in the spreadsheet:

links

If the puzzle URL is updated, the link in the spreadsheet will be updated as well.

For meta puzzles, feeder puzzles, answers, and links are prepopulated in a sheet name "AUTOGENERATED":
meta-feeders
The puzzles are sorted by answer.

Summary of changes:
* renamed GoogleDriveClient to GoogleApiClient
* added `add_puzzle_and_slack_links_to_sheet` and `__extract_id_from_sheets_url` methods
* renamed `GOOGLE_DRIVE_API_PRIVATE_KEY` to `GOOGLE_API_PRIVATE_KEY` and `GOOGLE_DRIVE_API_AUTHN_INFO ` to `GOOGLE_API_AUTHN_INFO`
* added logic to insert puzzle link and Slack channel link into spreadsheet on creation and when puzzle link is updated
* regenerate "AUTOGENERATED" sheet for all related metas if any of the following happens:
    * a puzzle is deleted
    * a puzzle is edited (name or URL changed)
    * a puzzle is tagged with a meta
    * a meta tag is removed from a puzzle
    * a puzzle is assigned to a meta
    * a puzzle is unassigned from a meta
    * an answer is marked CORRECT
    * a CORRECT answer's status is changed
* meta spreadsheet is updated in separate thread so we can still respond to the request quickly

For #134 and #74.


erwa:
Note to self: after deploying, need to update environment variables on Heroku.

google_api_lib/google_api_client.py
```diff
@@ -0,0 +1,214 @@
...
+        latest feeder puzzle info
+        '''
+        if not meta_puzzle.is_meta:
+            return
+
+        client = GoogleApiClient.getInstance()
+        if not client:
+            return
+
+        # update meta sheet in separate thread
```
daiweili: This is probably fine, but usually this would be handled with a task queue: https://djangopackages.org/grids/g/workers-queues-tasks/

Instructions for RQ / Celery on Heroku: https://devcenter.heroku.com/categories/background-jobs-in-python

google_api_lib/google_api_client.py
```diff
@@ -0,0 +1,214 @@
...
+            }]
+        }
+
+        self._sheets_service.spreadsheets().batchUpdate(
+            spreadsheetId=spreadsheet_id,
+            body=body
+        ).execute(http=http)
+
+    @staticmethod
+    def update_meta_sheet_feeders(meta_puzzle):
```
daiweili: One race condition with this:

Suppose 2 people make updates:
- Update 1 starts updating the spreadsheet
- Update 2 starts updating and finishes updating the spreadsheet
- Update 1 commits to the spreadsheet

The spreadsheet ends up in a stale final state until the next update. You could use redlock as a distributed lock to solve this: https://blog.heroku.com/heroku-redis-now-available-in-public-beta

google_api_lib/google_api_client.py
```diff
@@ -0,0 +1,98 @@
+import googleapiclient.discovery
```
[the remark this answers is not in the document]
erwa: Yeah, I used `git mv` but then made too many changes to the file so git treated it as a delete + add. Git has some heuristic to determine when something's a move and when it's a delete + add. Let me go back and do the git mv as one commit and the changes as a separate commit so it's more clear.

google_api_lib/google_api_client.py
erwa: Yeah, I saw a lot of mentions about that but didn't feel like introducing another library, learning it, and setting it up. Using threads seemed easier and more lightweight :)

erwa: Using RQ doesn't seem too difficult. I'll add a TODO to switch to that (or something else) later. Ref: https://devcenter.heroku.com/articles/python-rq

Though I think a DB-based queue is preferable since we're already running a DB (and RQ requires running a Redis instance). So maybe something like Django Background Tasks is preferable: https://django-background-tasks.readthedocs.io/en/latest/

google_api_lib/google_api_client.py
erwa: Lol, you really like Redis, huh?

You're right that there are race conditions and also we don't retry or rollback anything in case of failures midway. I think the proper way to solve these problems is to put this background work in a queue and use worker processes with some sort of locking (perhaps Redlock, perhaps something else). I'll add this to a TODO as well.

I think it's unlikely we'll encounter many race conditions in practice since on average, and even when they happen, the consequences aren't very dire -- we can always make a dummy change to retrigger updating the meta spreadsheet.

erwa: Another simple way of solving the race conditions is just to use a thread pool with 1 thread or when we migrate to background processes, a single worker process. 1 thread/process should be sufficient for our purposes.

google_api_lib/google_api_client.py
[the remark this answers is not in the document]
erwa: Ok, split my first commit into two commits (one just doing the `git mv`), so now the changes to GoogleApiClient should be clear:
* git mv commit: https://github.com/cardinalitypuzzles/smallboard/pull/140/commits/77098752ab59db24e0f95c459a6fd58bf917b931
* changes: https://github.com/cardinalitypuzzles/smallboard/pull/140/commits/224c02f6911090f2e4f9999f0635ea26bfbe2fdb

google_api_lib/google_api_client.py
erwa: Updated in https://github.com/cardinalitypuzzles/smallboard/pull/140/commits/24b088042c834f906e1ca33f61f2858822977752 to use a ThreadPoolExecutor with 1 worker.

erwa: Also added TODO to migrate to work queue and separate worker process in future.

daiweili: Ah, ok, this probably works fine for our deployment, but you will need locking with multiple Django instances or multiple worker queues.

Good point on retrying on failure too.

> I think it's unlikely we'll encounter many race conditions in practice since on average, and even when they happen, the consequences aren't very dire -- we can always make a dummy change to retrigger updating the meta spreadsheet.

Yeah, definitely not serious, and solvable with a workaround as you mention.

erwa: Oh, good point, I didn't think about the case of having multiple dyno instances, which we plan to have during the hunt.

I added a comment about needing locking: https://github.com/cardinalitypuzzles/smallboard/commit/ea27a7e6413291814300a11f7914b8bee0af78cd

PR #187, opened 2020-11-22
update python

rawxfish:
Updating to 3.6.12, which is the current heroku default: https://devcenter.heroku.com/articles/python-support


erwa (approved):
Any change we could upgrade to Python 3.7 or 3.8? There are some useful features in newer Python versions I sometimes like to use.

(Maybe outside the scope of this PR but would be nice to bump to something newer if it just works)

rawxfish:
Sure, I'm running it locally with 3.8 and it works fine, so it should be fine to bump it up.

PR #188, opened 2020-11-22
add privacy policy

rawxfish:
Google OAuth API has a verification process now that asks for a link to a privacy policy on our site: https://support.google.com/cloud/answer/9110914?hl=en

I generated one real quick from using terms feed. I think it just says we use emails?
Screen Shot 2020-11-22 at 5 51 12 PM


erwa (approved):
LGTM. The privacy page is loadable without being logged in, right?

rawxfish:
You do not have to be logged in to see it

Issue #183, opened 2020-07-17
shorten URL pasted into sheet


rawxfish:
It's annoying for sheet formatting to have a long url in the first column

PR #190, opened 2020-11-23
issue #183: shorten URL pasted into sheet

rawxfish:
Also gets rid of slack link

Screen Shot 2020-11-22 at 11 32 32 PM

### DIFF
--- a/google_api_lib/google_api_client.py
+++ b/google_api_lib/google_api_client.py
@@ -93,8 +93,6 @@ class GoogleApiClient:
         req_body = {
             'values': [
                 [f'=HYPERLINK("{puzzle_url}", "Puzzle Link")'],
-                #['Slack channel', '%s/app_redirect?channel=%s'
-                #                  % (settings.SLACK_BASE_URL, slack_channel_id)]
             ]
         }
         self._sheets_service.spreadsheets().values().update(
--- a/google_api_lib/google_api_client.py
+++ b/google_api_lib/google_api_client.py
@@ -92,15 +92,15 @@ class GoogleApiClient:
                                             sheet_url):
         req_body = {
             'values': [
-                ['Puzzle link', puzzle_url],
-                ['Slack channel', '%s/app_redirect?channel=%s'
-                                  % (settings.SLACK_BASE_URL, slack_channel_id)]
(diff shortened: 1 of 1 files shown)


google_api_lib/google_api_client.py
```diff
@@ -92,15 +92,15 @@ def add_puzzle_and_slack_links_to_sheet(self, puzzle_url, slack_channel_id,
                                             sheet_url):
         req_body = {
             'values': [
-                ['Puzzle link', puzzle_url],
-                ['Slack channel', '%s/app_redirect?channel=%s'
-                                  % (settings.SLACK_BASE_URL, slack_channel_id)]
+                [f'=HYPERLINK("{puzzle_url}", "Puzzle Link")'],
+                #['Slack channel', '%s/app_redirect?channel=%s'
+                #                  % (settings.SLACK_BASE_URL, slack_channel_id)]
```
akirabaruah, line 97: Why are these lines commented out? If we don't need them, we should probably just delete them.

akirabaruah (commented):
LGTM overall 👍  a bit curious about the commented lines though

google_api_lib/google_api_client.py
rawxfish: I was just being skittish about deleting code. I'll do it...

akirabaruah:
RIP slack. Did we want to remove all slack references from the codebases or just in this google sheet creation logic?

google_api_lib/google_api_client.py
akirabaruah: haha same, but I guess we've got git to back us up? 💪 

anyway, thanks for the cleanup!

akirabaruah (approved):

rawxfish:
Depends on how likely it is that discord becomes worse and we want to switch back to slack in the future... 
@asdfryan wrote all the slack integration last year, any thoughts how easy it would be to de- or re-tangle?

asdfryan:
> Depends on how likely it is that discord becomes worse and we want to switch back to slack in the future...
> @asdfryan wrote all the slack integration last year, any thoughts how easy it would be to de- or re-tangle?

YAGNI. Just delete it all IMO.

erwa (approved):
LGTM

Issue #156, opened 2020-01-15
Small Board should support multiple hunts


erwa:
A nice-to-have for the future is for Small Board to support multiple hunts. Currently, we need to wipe the database before each hunt.

In order to support multiple hunts, most of our database queries need to be made hunt-specific. Right now, some of our queries are hunt-specific (they specify hunt_pk) while others are not, which can lead to issues like the following:
```
Internal Server Error: /hunts/1/puzzles
Traceback (most recent call last):
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/venv/lib/python3.7/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/venv/lib/python3.7/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/hunts/views.py", line 91, in puzzles
    sorted_np_pairs = PuzzleTree(puzzle_objects).get_sorted_node_parent_pairs()
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/puzzles/puzzle_tree.py", line 77, in __init__
    node.parents = [self.puzzle_to_node[meta.pk] for meta in p.metas.all()]
  File "/mnt/c/Users/volde/Desktop/github/cardinalitypuzzles/smallboard/puzzles/puzzle_tree.py", line 77, in <listcomp>
    node.parents = [self.puzzle_to_node[meta.pk] for meta in p.metas.all()]
[...]

PR #193, opened 2020-11-24
Some changes for supporting multiple hunts

rgossiaux:
* Puzzle names should only be unique per-hunt
* The old channel/sheets defaults when integrations are off needed to be
  updated to use unique values; make the fields nullable and use null in
  those cases (makes more sense anyway imo)
* Fix a few queries


rgossiaux:
Some context in #156; don't think this is enough to make multiple hunts work perfectly, just an incremental improvement

hunts/views.py
```diff
@@ -162,8 +162,8 @@ def post(self, request, pk):
             if google_api_client:
                 sheet = google_api_client.create_google_sheets(name)
             else:
-                # TODO(erwa): This should incur a warning.
```
akirabaruah, line 165: Did @erwa mean a warning log or a user-visible warning on the site? If the latter, we should probably leave the TODO. Log still seems useful though.

puzzles/models.py
```diff
@@ -58,6 +58,11 @@ class Puzzle(models.Model):
 
     active_users = models.ManyToManyField('accounts.Puzzler', related_name="active_puzzles")
 
+    class Meta:
+        constraints = [
+                models.UniqueConstraint(fields=['name', 'hunt'], name='unique_names'),
```
akirabaruah, line 63: nit: Consider "unique_names_per_hunt" - if it seems too verbose, don't worry about it.

puzzles/models.py
```diff
@@ -27,12 +27,12 @@ class InvalidMetaPuzzleError(PuzzleModelError):
...
 
 class Puzzle(models.Model):
-    name = models.CharField(max_length=80, unique=True)
+    name = models.CharField(max_length=80)
     hunt = models.ForeignKey('hunts.Hunt', on_delete=models.CASCADE, related_name='puzzles')
     url = models.URLField(blank=True)
 
-    sheet = models.URLField(default='', unique=True)
-    channel = models.CharField(max_length=128, default='', unique=True)
+    sheet = models.URLField(default='', unique=True, null=True)
```
akirabaruah, line 34: Should we replace `default=''` with `blank=True` for readability?

hunts/views.py
```diff
@@ -162,8 +162,8 @@ def post(self, request, pk):
...
-                sheet = puzzle_url
+                logger.warn("Sheet not created for puzzle %s" % name)
+                sheet = None
```
akirabaruah, line 166: Do we need to update any templates to account for this? I recall that the site defaulted to the puzzle URL when a sheet wasn't available, so I'm wondering if this might break something here: https://github.com/cardinalitypuzzles/smallboard/blob/cba90c7c6fc8490727895695b9083ec26fa87a2e/hunts/templates/all_puzzles.html#L347

akirabaruah (commented):
We don't really have tests yet, do we? :sweat_smile:

akirabaruah:
Thanks for the quick PR @rgossiaux!

puzzles/models.py
```diff
@@ -27,12 +27,12 @@ class InvalidMetaPuzzleError(PuzzleModelError):
...
+    channel = models.CharField(max_length=128, default='', unique=True, null=True)
```
asdfryan, line 35: Now that we are allowing null, should the default be None?

Same question for the sheet URL.

akirabaruah: How about having `(unique=True, null=True, blank=True)` to include proper form validation (and default semantics) as described here? https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.Field.null

rgossiaux: Seems like the right thing is both changing `default=None` (an empty string default violates the uniqueness constraint anyway) and adding `blank=True`(not sure when this would come up but maybe in the admin console)

hunts/views.py
rgossiaux: I tested this-- I think the front-end already handles this correctly: 
image

hunts/views.py
rgossiaux: Not sure; personally I'm fine with no user warning (if smallboard is being run without google credentials it's going to be obvious anyway since there won't be any sheets created) but open to other opinions

hunts/views.py
akirabaruah: Ah perfect - thanks for checking!

hunts/views.py
```diff
@@ -162,8 +162,8 @@ def post(self, request, pk):
             if google_api_client:
                 sheet = google_api_client.create_google_sheets(name)
             else:
-                # TODO(erwa): This should incur a warning.
-                sheet = puzzle_url
```
erwa, line 166: nice, None seems more appropriate

erwa (approved):
LGTM

puzzles/models.py
akirabaruah: I think this question is covered in the comment thread below: https://github.com/cardinalitypuzzles/smallboard/pull/193/files#r529159104

PR #198, opened 2020-11-24
Add pre-commit checks

akirabaruah:
Installs the pre-commit tool and adds checks including:

- black python formatter: https://github.com/psf/black
- ~~conventional commits: https://www.conventionalcommits.org/en/v1.0.0/~~

Also applies these formatters to the existing codebase for consistency.


rgossiaux:
python autoformatting sounds good to me. should we ignore migrations?

personally not a big fan of conventional commits tbh, not sure i see the benefit

akirabaruah:
> python autoformatting sounds good to me. should we ignore migrations?

Is there something about migrations that should prevent us from formatting? I figure that migrations are also python code, so we might as well run the formatter on them too for consistency. It's also less to configure and one less exception to worry about :+1: 

> personally not a big fan of conventional commits tbh, not sure i see the benefit

Seems like the biggest wins would be:

- consistency in commit history to improve readability
- standardized format for backwards-incompatible changes
- automatic changelog generation

(see https://www.conventionalcommits.org/en/v1.0.0/#why-use-conventional-commits)

In general, I'm just hoping to have some sort of formatting guidelines (and linter) for commits to make it easier to navigate the change history. Doesn't have to be conventional commits, but that system was the simplest and most documented solution I could find :smile: any alternative suggestions are welcome!

akirabaruah:
I'm also fine with removing the commit message hooks for now, since I can see that it might add some extra friction for devs getting started. But it would be great if we can agree on some sort of commit message styling. Thoughts?

akirabaruah:
@rgossiaux I removed the conventional commits linter as you suggested, since it'll will probably incur more dev headache than it will provide benefit. Left the revert commits in here in case we ever decide to implement it later.

Re: formatting migrations, it's possible that devs might have to tweak the django-generated code as models become more complex, so for consistency, I've left black configured to include migration code. Lemme know if you disagree.

erwa (approved):
+1 for enforced style

PR #199, opened 2020-11-24
Config sqlite env for local development

akirabaruah:
Simplifies local development by avoiding the postgresql requirement since the sqlite3 database backend is built into Django.


rawxfish (approved):

akirabaruah:
@rgossiaux FYI since you're working on a docker-based env

erwa:
Nice, thanks for simplifying local setup. Postgres was a bit of a headache -- setup doesn't quite work the same on all platforms. The comes-with-Django Sqlite seems better.

Issue #197, opened 2020-11-24
Complete Google OAuth verification


rawxfish:
We use google oauth to make sure only people with access to the drive can access smallboard.

https://support.google.com/cloud/answer/9110914?hl=en

I added a privacy page but apparently our home page also needs work:

![image](https://user-images.githubusercontent.com/1312469/100050302-1fd9e100-2de7-11eb-8118-6990e8bb798e.png)

akirabaruah:
Sounds like we might want to add a generic home page that explains what smallboard is and link to the privacy policy?

rawxfish:
We got a followup email today:

![image](https://user-images.githubusercontent.com/1312469/100827253-51c3f680-342a-11eb-97ee-d7cada8c16f0.png)

Do you guys think this is sufficient for the home page:

![Screen Shot 2020-12-01 at 11 11 48 PM](https://user-images.githubusercontent.com/1312469/100827399-9bacdc80-342a-11eb-803c-a88ec343ff6e.png)

akirabaruah:
LGTM overall - thanks for keeping tabs on this - a couple things we might want to clarify:

- the whitelist of emails ensures that only authorized users access the site (as an aside, do we want to eventually open it up to the public?)
- under "How do I join a hunt?" - emailing us will only let you access this app, not actually add you to <insert puzzle hunt name here>

rgossiaux:
Maybe we could mark it internal-only for now? Do we qualify for that? Might be easier than wrangling with them.

rawxfish:
I think internal-only means setting up an organization and giving everyone an org email: https://support.google.com/cloud/answer/6158849#public-and-internal

rawxfish:
> the whitelist of emails ensures that only authorized users access the site (as an aside, do we want to eventually open it up to the public?)

I'm worried that if the language isn't public facing, they'd make us be an internal app and pay money to be an org...

> under "How do I join a hunt?" - emailing us will only let you access this app, not actually add you to

How about "How do I get whitelisted?"

akirabaruah:
Yeah, at this point, do you think it's easier to put up this new homepage vs. set up individual user emails under a private org?

Depending on how much/little Google nit-picks your proposed homepage, it seems like we're almost there in terms of being a public Google-approved app?

akirabaruah:
> I'm worried that if the language isn't public facing, they'd make us be an internal app and pay money to be an org...

[...]

Issue #191, opened 2020-11-24
Multi-tenancy: smallboard should support per-hunt settings to allow multiple teams to use the same smallboard instance simultaneously


akirabaruah:
Right now, Smallboard assumes a single team (Cardinality), but it'd be great to support users creating different teams and working on hunts using the same site in the long run.

See https://github.com/cardinalitypuzzles/smallboard/issues/156#issuecomment-732500049 for prior discussion.

akirabaruah:
This is probably not a high priority for Cardinality's prep for the upcoming MIT Mystery Hunt 2021. But I'm sure there are lots of puzzlers around the world who might enjoy using this tool for other teams, so it's probably worth tracking for some milestone.

akirabaruah:
One change to make this work is the association between Puzzles and Answers, since different teams will likely work on the same Puzzles with different Answers, at different solve rates. We'd probably want to use the existing `Puzzle` model as the unique representation of a given puzzle and introduce something like a `SolveAttempt` model relating a Team's Answers to a particular Puzzle.

rgossiaux:
It might be easiest to just sandbox every hunt from each other, so each hunt instance corresponds to one team's view of one hunt and nothing is shared between hunt instances

akirabaruah:
I've been itching to get this working, so I'm planning on putting up a WIP PR this week 👍

erwa:
Copying over a comment from #501:

> Currently any user who can login to a smallboard instance can see all the created hunt instances.
>
> It would be nice if several different teams could use the same smallboard instance for the same and/or different hunts simultaneously, where each team creates its own hunt instance specifying its own Google Drive folder, team members who have access, Sheets template file, Discord server, etc.

erwa:
Note an initial stab at per-hunt settings was done in #538.

rgossiaux:
I think the biggest piece still missing here is per-hunt permissions. Login right now is still tied to one global Drive folder & we don't have a way of separating the permissions for individual hunts. I think we probably want to add something like [django-guardian](https://django-guardian.readthedocs.io/en/stable/overview.html) and then integrate that with the API. It's probably not *too* much work once somebody finds the time to actually do it.

The other part is actually exposing a way of configuring these settings in the UI--some kind of hunt settings page. This can probably just use a Django view/form, it doesn't need React. We can use the object permissions from above to restrict this page to the person that created the hunt (and if we're feeling fancy, anyone else they delegate).

PR #201, opened 2020-11-26
Deprecate ACTIVE_HUNT_ID env variable

akirabaruah:
Rather than relying on an app environment variable, we can use the
boolean "active" field on Hunt objects in the database to find the
current active hunt.

Removes all uses of ACTIVE_HUNT_ID and updates dependent logic to use a
database query for active=True instead.

This will help in supporting true multi-hunt functionality (see https://github.com/cardinalitypuzzles/smallboard/issues/156).


akirabaruah:
@erwa I think you originally added ACTIVE_HUNT_ID, so I added you as a reviewer (other folks, feel free to review as well :smile:)

hunts/views.py
```diff
@@ -214,5 +216,20 @@ def post(self, request, pk):
         return JsonResponse({'data': result})
 
 
+class ActiveHuntRedirectView(RedirectView):
+    pattern_name = 'hunts:all_puzzles'
+
+    def get_redirect_url(self, *args, **kwargs):
+        try:
+            active_hunt = Hunt.objects.get(active=True)
```
rgossiaux, line 224: I think this is going to raise an exception when there are multiple hunts marked active (my local db is in this state for example).

What do you think about keeping track of it per-user? Maybe it's overkill but we could maybe then update it on each page load, so whatever hunt you were last looking at is your default.

rgossiaux:
thanks for getting rid of this, it helps a lot with setup/config imo :)

hunts/views.py
akirabaruah: Ooh yeah that sounds good. So, if the user's never opened smallboard we can load a generic home page (probably needed for https://github.com/cardinalitypuzzles/smallboard/issues/197 anyway as @rawxfish pointed out), otherwise redirect to the last accessed hunt. To achieve this, I'm thinking we'll want to:

- add an `active_hunt = ForeignKey(Hunt, null=True)` to the `Puzzler` user model
- update `Puzzler.active_hunt` whenever a user accesses a `Hunt`
- redirect based on `Puzzler.active_hunt` or load the home page if it's null
- remove the `Hunt.active` field

Am I missing anything?

erwa:
> @erwa I think you originally added ACTIVE_HUNT_ID, so I added you as a reviewer (other folks, feel free to review as well 😄)

Actually, @rawxfish originally added it in 914128a5 :).

smallboard/urls.py
```diff
@@ -19,15 +19,16 @@
...
-    path('answers/', include('answers.urls')),
-    path('privacy', views.privacy, name='privacy'),
-    path('', lambda r: HttpResponseRedirect('hunts/{}/'.format(settings.ACTIVE_HUNT_ID))),
-    path('', include('social_django.urls', namespace='social')),
+    path("admin/", admin.site.urls),
+    path("accounts/", include("django.contrib.auth.urls")),
+    path("accounts/", include("accounts.urls")),
+    path("puzzles/", include("puzzles.urls")),
+    path("hunts/", include("hunts.urls", namespace="hunts")),
+    path("answers/", include("answers.urls")),
```
erwa, line 30: Is the namespace="hunts" part necessary? Or else, should we also add namespace="answers" for consistency?

hunts/views.py
erwa: Some related thoughts: It'd be nice if there's somewhere users can toggle between hunts if they are working on multiple hunts simultaneously.

Also, we might consider having unique URLs per hunt so you can differentiate hunts just by looking at that. E.g.: hunt/mitmh2021/... vs. hunt/crums2020/... etc.

Another use case to think about supporting in the future (which has been discussed elsewhere) is having multiple instances of the same hunt, so different teams can use smallboard for the same hunt and not affect each other or see each other's answers. Maybe this could be accomplished by introducing the notion of a Team and having puzzlers and answer submissions associated with teams. Just something to think about.

erwa (commented):
This looks like a step in the right direction. Have a few comments and questions.

hunts/views.py
akirabaruah: > Some related thoughts: It'd be nice if there's somewhere users can toggle between hunts if they are working on multiple hunts simultaneously.

Maybe something like a user settings page? Filed https://github.com/cardinalitypuzzles/smallboard/issues/203

> Also, we might consider having unique URLs per hunt so you can differentiate hunts just by looking at that. E.g.: hunt/mitmh2021/... vs. hunt/crums2020/... etc.

Do you mean a site URL per hunt? In other words, maybe a separate page per hunt?

Or do you mean enforcing a unique *external* URL per hunt?

> Another use case to think about supporting in the future (which has been discussed elsewhere) is having multiple instances of the same hunt, so different teams can use smallboard for the same hunt and not affect each other or see each other's answers. Maybe this could be accomplished by introducing the notion of a Team and having puzzlers and answer submissions associated with teams. Just something to think about.

:+1: My thoughts exactly (and @rgossiaux's too, I think). See https://github.com/cardinalitypuzzles/smallboard/issues/191 for the "elsewhere" discussions :smile:

Are any changes we should make in this PR to address the different instances per Team?

rgossiaux: @akirabaruah those changes sound good to me, though you could probably leave Hunt.active as it is if you wanted; not really sure about the context but I'm guessing the idea was to distinguish between hunts in-progress and completed hunts which is a little different from "which hunt a user is/was working on".

Really like the unique URL per hunt, that's a lot more readable than /2/ and stuff lol

akirabaruah:
Added a slug field for hunts so that they're human readable. I ended up leaving the Hunt.active field as @rgossiaux suggested and added a last_accessed_hunt to the User model to clarify the field's intended purpose.

akirabaruah (commented):
Added a slug field for hunts so that they're human readable. I ended up leaving the Hunt.active field as @rgossiaux suggested and added a last_accessed_hunt to the User model to clarify the field's intended purpose.

answers/urls.py
```diff
@@ -3,10 +3,13 @@
...
 from django.urls import path
 
+from hunts.views import LastAccessedHuntRedirectView
+
+app_name = 'answers'
 urlpatterns = [
-    path("queue/<int:hunt_pk>", AnswerView.as_view()),
-    path("queue/<int:hunt_pk>/answers", answers),
-    path("queue/<int:hunt_pk>/<int:answer_pk>", AnswerView.as_view()),
+    path("queue/<slug:hunt_slug>", AnswerView.as_view(), name="hunt_queue"),
```
erwa, line 10: How about `submission_queue` for the name?

erwa (approved):
I like the namespacing of URLs -- makes it more clear what the URLs are for. Also, thanks for adding the hunt slug field!

hunts/views.py
```diff
@@ -214,5 +216,23 @@ def post(self, request, pk):
...
+        hunt = self.request.user.last_accessed_hunt
+        if not hunt:
+            return reverse('hunts:index')
+
+        # Answer urls use pattern_name="answers:hunt_queue" and "hunt_slug" to
+        # disambiguate between answer and hunt primary keys.
+        if self.pattern_name == LastAccessedHuntRedirectView.pattern_name:
+            kwargs['slug'] = hunt.slug
+        else:
+            kwargs['hunt_slug'] = hunt.slug
```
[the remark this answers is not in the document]
akirabaruah: As far as I can tell, within the `hunt` app, it was referred to as "id" whereas in the `answers` app, it was called "hunt_id" to disambiguate from Answer "id" values. "hunt_id" / "hunt_slug" seems *slightly* redundant within the `hunt` app, but it seems worth using everywhere to reduce this RedirectView's complexity :+1:

akirabaruah:
I think I've incorporated everyone's feedback so far (barring anything we deferred to newly filed Issues. Please let me know if I've missed anything - thanks!

Tests to follow :smile:

akirabaruah:
> Tests to follow smile

Actually, I'm gonna merge this as is, then follow up with tests after getting some coverage metrics. Otherwise, it's a bit unclear when we've added a sufficient level of testing.

PR #202, opened 2020-11-26
Replace 'accounts.Puzzler' with get_user_model()

akirabaruah:
ForeignKey/ManyToManyFields in other app Models are more generalized and
scalable this way.


akirabaruah:
This is just a simple cleanup :broom:

erwa (approved):
Nice, this seems better than hardcoded strings.

PR #208, opened 2020-11-29
Add Docker support

rgossiaux:
Running `docker-compose up` in the project root will now start smallboard.

I left some of the non-docker instructions where they are if people want to keep using their existing environments locally & don't want to switch to using docker at the moment.


docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      - POSTGRES_DB=smallboard
+  web:
+    build: .
+    command: python manage.py runserver 0.0.0.0:8000
+    volumes:
+      - .:/usr/src/smallboard/
+    ports:
+      - 8000:8000
+    env_file:
+      - ./.env.docker
```
akirabaruah, line 20: Why not provide environment variables via `environment` as for the `db` container above? That way, all of the configuration can stay in this one file.

`.env` seems to be a convenience for local development, but shouldn't we be using env vars for production-like use cases like Docker?

akirabaruah (commented):
Yay Docker :tada: Do we need to rebase these changes?

docker-compose.yml
rgossiaux: Lol that's a much better idea (you can tell this is the first docker config I've written)

docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - POSTGRES_DB=smallboard
+  web:
+    build: .
+    command: python manage.py runserver 0.0.0.0:8000
+    volumes:
+      - .:/usr/src/smallboard/
+    ports:
+      - 8000:8000
+    environment:
+      - DATABASE_URL=postgres://smallboard_user:smallboard_password@db/smallboard
```
erwa, line 20: Is it possible to define the variables user, pw, and db in one place and then refer to them later, instead of duplicating the values? I forget if YAML lets you do this.

erwa (commented):
Nice!

It would be great if we could add building the Docker image as a check that's run for every PR so we don't accidentally break it.

docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    build: .
+    command: python manage.py runserver 0.0.0.0:8000
+    volumes:
+      - .:/usr/src/smallboard/:Z
+    ports:
+      - 8000:8000
+    environment:
+      - DATABASE_URL=postgres://smallboard_user:smallboard_password@db/smallboard
+    env_file:
+      - ./.env
```
akirabaruah, line 22: What do you think of just removing this `env_file` entirely? Now that we configure `DATABASE_URL` in `environment` the only other setting is `ACTIVE_HUNT_ID`, which will be deprecated by https://github.com/cardinalitypuzzles/smallboard/pull/201

Maybe more importantly, it reduces the possibility of configuring env vars in two separate places. For `docker-compose` it might make more sense to just use `environment` and complete ignore the `env_file`. Any local tweaks to the environment can be made in the `docker-compose.yaml` file.

docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: '3.8'
```
akirabaruah, line 1: nit: we should probably name this file `docker-compose.yaml` as per https://yaml.org/faq.html

docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    image: postgres:13.1-alpine
+    volumes:
+      - postgres_data:/var/lib/postgresql/data
+    environment:
+      - POSTGRES_USER=smallboard_user
+      - POSTGRES_PASSWORD=smallboard_password
+      - POSTGRES_DB=smallboard
+  web:
+    build: .
+    command: python manage.py runserver 0.0.0.0:8000
```
akirabaruah, line 14: We should probably include the `manage.py migrate` command in here to ensure that this works out of the box. Otherwise, simply running `docker-compose up` on a fresh repo will fail due to unapplied migrations. The `migrate` command is idempotent, so it should be safe to just tack on here.

e.g.

```
command: sh -c "python manage.py migrate && python manage.py runserver ..."
```

akirabaruah (changes requested):
Added a commit to include Z mounts for selinux systems :+1: this PR failed on my Fedora machine without it. Lemme know if it breaks for you?

docker-compose.yml
akirabaruah: I couldn't tell! It seems to work great for me (with some minor tweaks)

docker-compose.yml
akirabaruah: Seems like we can use extension fields or an env_file: https://stackoverflow.com/questions/36283908/re-using-environment-variables-in-docker-compose-yml

That way, `DATABASE_URL` can be constructed using the `POSTGRES_` variables and all of those variables can be defined globally to the compose file:

    POSTGRES_USER=smallboard_user
    POSTGRES_PASSWORD=smallboard_password
    POSTGRES_DB=smallboard
    DATABASE_URL="postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db/${POSTGRES_DB}"

I'd advocate for *either* storing all variables in the docker-compose file using extension fields or storing all variables in a `docker.env` file, but not both. The latter option seems a bit simpler in this case tbh.

docker-compose.yml
erwa: Loading some environment variables from an env_file might still be nice because there are some environment variables that we don't want to check in, like the GOOGLE_API_PRIVATE_KEY and later, the Discord API key.

docker-compose.yml
rgossiaux: Yeah I was on the fence about this one, I think it might be a little unexpected if we ran migrations automatically locally and maybe could do something bad if you ran this in the middle of making a change. On the flipside that might not be a big deal. I think if we ended up wanting to deploy with this we would definitely want to make sure it ran though.

docker-compose.yml
rgossiaux: Right, I think we still want this for all our secrets and stuff.

rgossiaux:
Nice, thanks for testing and fixing on Linux, glad to know it actually works there lol

docker-compose.yml
rgossiaux: Hmm it doesn't look to me like you can use extension fields to construct the DATABASE_URL. 

I think the snippet you posted requires setting those env variables on the host machine, right? Not sure this is really worth it, would rather just keep the config simpler. I don't think the duplication is a big deal.

docker-compose.yml
rgossiaux: Ok, decided to just do it (and run collectstatic as well)

docker-compose.yml
erwa: Yeah, it seems YAML doesn't have quite the substitution support we needed.

Agreed it's not a big deal and we can just leave it as is for now.

erwa (approved):

dev-guide.md
```diff
@@ -2,14 +2,19 @@
 
 ### Getting started
 
-To set up Small Board locally, you need:
+There are two ways to set up Small Board locally.
+
+To run Small Board in Docker, you will need:
+
+* Git
+* Docker (on Mac or Windows, install Docker Desktop; on Linux, docker and docker-compose)
```
akirabaruah, line 10: nit: Should we include a link to the docker downloads page?

docker-entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
akirabaruah, line 1: ooh nice!

dev-guide.md
```diff
@@ -111,7 +130,27 @@ source venv_smallboard/bin/activate
...
-#### Copying production data to local database
+#### Local deployment
+
+Once the Python environment and database are set up and running, you can run Small Board locally using
+
+```
+(venv_smallboard)$ python manage.py runserver
+```
+
+You can view the app in your browser at [http://127.0.0.1:8000/]().
```
akirabaruah, line 141: Maybe we can use the markdown shortcut `<http://...>` for these kinds of links? https://daringfireball.net/projects/markdown/syntax#autolink

akirabaruah (approved):
LGTM with some nits :+1: thanks for making this happen! I think it'll help enable even more developers

dev-guide.md
rgossiaux: Sure (I didn't change this, just reshuffled it a bit)

docker-compose.yml
akirabaruah: Ah just saw this, but yeah I agree with your assessment of the tradeoffs 👍 I guess for development, we can encourage just spinning up the docker compose once, then making changes and having the running django instance hot-reload changes as usual (assuming we've mounted the dev's local code into the container volume).

PR #210, opened 2020-11-29
Only strip whitespace from answers

### LINKED ISSUE #168: Small Board should support emojis
At the MITMH 2020, there were some answers that were emojis. Small Board currently strips all non A-Z or 0-9 characters:
* https://github.com/cardinalitypuzzles/smallboard/blob/master/puzzles/templates/modals/submit_answer.html#L42-L44
* https://github.com/cardinalitypuzzles/smallboard/blob/master/puzzles/views.py#L49-L50

We may want to allow emojis for future hunts.

rgossiaux:
Fixes #168.

Deleted the validation on the JS-side since it's not needed.


akirabaruah:
Nice :+1: LGTM

akirabaruah (approved):

rawxfish (approved):

erwa (approved):

PR #211, opened 2020-11-29
Redraw expander arrows when changing the table

### LINKED ISSUE #119: Expand/collapse triangles disappear after Search
Initially, you see triangles:
initial

When you start searching, the triangles disappear:
search

When you clear the search box, triangles are still gone:
clear

However, functionality still works -- you can click the white space where the triangle should be and collapse still works. Also, when you click, the triangle reappears:
reappear

@asdfryan FYI

### LINKED ISSUE #159: collapse meta button is missing on page load
When the page first loads, the collapse button is missing but still clickable. It shows up after you collapse and uncollapse though.
Screen Shot 2020-01-15 at 10 32 44 PM

rgossiaux:
Fixes #159.
Fixes #119.
The initTreegrid function seems to blow away the expander arrows
somehow. Telling it to rerender those whenever we recreate the treegrid
seems to do the trick.

Also moved some code to set callbacks to a better place.

### DIFF
--- a/hunts/templates/all_puzzles.html
+++ b/hunts/templates/all_puzzles.html
@@ -224,7 +224,7 @@ function reload() {
 }
 
 var all_uncollapsed = false;
-function maybeExpandAll() {
+function onSearch() {
     if ( all_uncollapsed == false ) {
         all_uncollapsed = true;
         $('.tree').treegrid('expandAll');
@@ -237,6 +237,7 @@ function initTreegrid() {
             all_uncollapsed = false;
         }
     });
+    $('.tree').treegrid('renderExpander');
     if ($('.tree').hasClass( "initial-collapsed" )) {
         $('.initial-collapsed').treegrid('collapse');
     }
@@ -323,11 +324,10 @@ $(document).ready(function() {
             $(row).attr('id', `puzzle-${data[pk]}`);
             initDeleteTagHandlers(data);
         },
-        'drawCallback': function() {
-            $('#puzzles').on('draw.dt', initTreegrid);
-            $('#puzzles').on('search.dt', maybeExpandAll);
-        },
         'initComplete': function() {
+            initTreegrid();
+            $('#puzzles').on('search.dt', onSearch);
+            $('#puzzles').on('draw.dt', initTreegrid);
             reload();
         },
         'columnDefs': [


asdfryan:
Does re-rendering add latency? How does it perform on 50+ puzzles?

I remember Daiwei made a bunch of optimizations because we had a huge latency issue last year. (and had some unfortunate side effects, like getting rid of the expander arrows)

rgossiaux:
This change only adds calls to `initExpanders` which is just a simple function (takes 0.3 ms in a profile I just ran).

In fact I think this is actually a very large performance boost-- the old `drawCallback` was called every time the table redrew and added tons of duplicate event handlers calling `initTreegrid`, which is very expensive (50-100+ ms for a single call with only a dozen puzzles). When I test on master just loading the page and typing in and backspacing a quick search, I already see four handlers attached, and as the window stayed open and changes were made, this would only increase. I wouldn't be surprised if this was a main source of latency last year.

hunts/templates/all_puzzles.html
```diff
@@ -323,11 +324,10 @@ <h5 class="modal-title" id="addtitle">Add Puzzle</h5>
...
-        'drawCallback': function() {
-            $('#puzzles').on('draw.dt', initTreegrid);
-            $('#puzzles').on('search.dt', maybeExpandAll);
-        },
         'initComplete': function() {
+            initTreegrid();
+            $('#puzzles').on('search.dt', onSearch);
+            $('#puzzles').on('draw.dt', initTreegrid);
             reload();
         },
```
erwa, line 332: I remember needing drawCallback so when the data got re-polled every 5-10 seconds, the table would automatically refresh. After your changes, if you enter an answer or add a puzzle in a different browser tab, will the table in the first tab still automatically refresh/redraw?

rgossiaux: Yep, still works. Every draw calls `drawCallback` and fires the `draw.dt` event. We just want exactly one listener to this event (the `$('#puzzles').on('draw.dt', initTreegrid)` call), but the existing code adds a new additional listener on every draw (so after n draws, there are n listeners).

erwa: Nice! Glad someone understand JS better than I do :D

erwa (approved):

asdfryan (approved):

PR #212, opened 2020-11-29
Add pre-commit github actions workflow

akirabaruah:
Automatically checks pre-commit linters on pull requests and commits to
the master branch. This actually enforces the pre-commit changes added
in https://github.com/cardinalitypuzzles/smallboard/pull/198.

See https://github.com/pre-commit/action for docs on the action used here.


erwa (approved):
Nice!

PR #230, opened 2020-11-30
move tools page code

rawxfish:
Tools page feels smallboard-wide rather than hunt-specific to me.


erwa (approved):
I agree that tools should be site-wide rather than hunt-wide.

One thing though: we threw some hunt-specific info like hunt HQ contact on the tools page, too. Might be useful if there can also be a per-hunt-specific page where hunt-specific info can be put? (we could do this in a separate PR, too)

akirabaruah:
Yeah maybe we can add something like a "hunt info" page?

This PR makes a lot of sense though :+1: thanks!

akirabaruah (approved):

PR #235, opened 2020-12-01
add Pipfile and Pipfile.lock

### LINKED ISSUE #194: Use pipenv instead of virtualenv
[Pipenv](https://pypi.org/project/pipenv/) has a few nice features over virtualenv like separating dev and prod Python dependencies. Pipenv can still generate a `requirements.txt` file if we want and Heroku supports it by default. It might be nice to transition to that tool and update the developer docs to use it instead of virtualenv.

cs-miller:
remove requirements.txt in favor of Pipfile/Pipfile.lock (https://pypi.org/project/pipenv/)
fixes #194


Pipfile
```diff
@@ -0,0 +1,24 @@
...
+django-heroku = "*"
+google-api-python-client = "*"
+social-auth-app-django = "*"
+django-taggit = "*"
+url-normalize = "*"
+
+[dev-packages]
+pre-commit = "*"
+black = "*"
+coverage = "*"
```
erwa, line 18: Is this Pipfile auto-generated? Can these packages be alphabetized?

erwa (commented):
This looks good to me overall. Just a heads up, you may need to change the .circleci/config.yml (or the .github/workflows/pre-commit.yaml file after #233) since they both currently do pip install -r requirements.txt.

Pipfile
cs-miller: Yeah, it's autogenerated. For whatever reason, running pipenv install <x> does not alphabetize the Pipfile.
I will look into a solution for this.

cs-miller: I'm planning to update the dockerfile and ci before landing this

cs-miller:
At the very least, I'm gonna wait on #233 so that I can update those config files

Pipfile
rgossiaux: +1 on updating the Docker config :)

rgossiaux:
just curious, have you looked into poetry at all? Looks like another popular tool for package management. i don't really have much experience with either one personally

cs-miller:
Yeah, I’m gonna take a look at it today

akirabaruah:
> just curious, have you looked into poetry at all? Looks like another popular tool for package management. i don't really have much experience with either one personally

to add on, @cs-miller and I chatted and had some of these thoughts last night:
1. Heroku seems to support `requirements.txt` (`virtualenv`) or `Pipfile` (`pipenv`) out of the box (no config required). Likely not the case for poetry, but also not a huge deal to configure it accordingly.
2. poetry follows the `pyproject.toml` standard format, which is also used by `black` for config.

dev-guide.md
```diff
@@ -56,23 +56,20 @@ Our docker setup reads environmental variables from .env as described down the p
...
+Once you have pipenv, you can create an environment by running:
 ```
-source venv_smallboard/bin/activate
-(venv_smallboard)$ pip install -r requirements.txt
+pipenv install --dev
+```
+This creates a virtual environment and installs the dependencies and dev dependencies
+You can then activate the virtual environment by running
+```
+pipenv shell
```
akirabaruah, line 71: We should update the docs below about calling `./manage.py ...`

(you already had plans for this, but commenting just for the record)

akirabaruah:
> At the very least, I'm gonna wait on #233 so that I can update those config files

FYI this has landed :+1:

cs-miller:
poetry seems to... not work? with fresh everything on WSL, so I'm not inclined to look into it very deeply.

cs-miller:
I think I should probably update the dev guide a bit more, but as it stands everything "works"

cs-miller:
i will say that its not super clear to me that this is a large improvement over the status quo, so i'm fine not merging this at all

akirabaruah:
The separation of direct deps from transitive deps seems like a win IMO. With `requirements.txt`, didn't we find some dangling deps that weren't cleaned up when we removed some other main deps that brought them in indirectly?

Plus, you already put in the effort to get this working, so I'd advocate for merging this PR 😄

cs-miller:
problem is it breaks people's existing workflow (must use pipenv). anyway, ill merge it if someone approves :)

akirabaruah (approved):
LGTM, but did we want to update the docs about using `./manage.py` commands with `pipenv`? Or at least remove references to `requirements.txt`?

akirabaruah:
> problem is it breaks people's existing workflow (must use pipenv). anyway, ill merge it if someone approves :)

yeah, I think as long as we update the dev guide docs accordingly, it should be okay?

rgossiaux:
> poetry seems to... not work? with fresh everything on WSL, so I'm not inclined to look into it very deeply.

Yeah seems fine, no need to look into it further. The Heroku support thing alone seems like a pretty compelling reason to use pipenv.

dev-guide.md
cs-miller: is this actually different from running whatever command is already in the docs (provided one is inside the env/shell)?

dev-guide.md
```diff
@@ -12,6 +12,7 @@ To run Small Board in Docker, you will need:
 To run Small Board manually, you will need:
 
 - Git
+- Python
```
akirabaruah, line 15: Should we include a minimum version? Or at least say Python 3?

dev-guide.md
```diff
@@ -121,18 +128,18 @@ Once this is set up, you'll need to run a one-time database migration to set up
...
+(smallboard-XYZ)$ python manage.py migrate
 ```
 
 #### Local deployment
 
 Once the Python environment and database are set up and running, you can run Small Board locally using
 
 ```
-(venv_smallboard)$ python manage.py runserver
+(smallboard-XYZ)$ python manage.py runserver
```
akirabaruah, line 142: Hmm, actually, maybe we should just remove this virtual env parens stuff and just say that all following commands assume that you're inside a `pipenv shell` environment? It will save us from having to add extra text to everything.

I also realized there's a bunch of commands below like `heroku` (should we add this to dev deps?) and `pre-commit` that rely on the environment.

akirabaruah (commented):
Still LGTM 👍

dev-guide.md
cs-miller: yeah im mixed on this, it wasn't consistent to begin with. i will remove

dev-guide.md
cs-miller: here i don't think it actually matters? you just need it to install pipenv, and i think most methods of installing pipenv actually pull in python if you don't have it already. i just included it here for completeness

PR #237, opened 2020-12-02
Start rewriting the front-end in React

rgossiaux:
This branch contains the work I've been doing the last 48 hours or so to write a new front-end for smallboard using React. I'm still working on it but I want to share it right now to open it up for comment, to let other people get involved, to make people aware I am doing this, and to try to merge things incrementally.

**Why do this?** The current front-end works pretty well and I don't want to rush into replacing it until the alternative is stable and mature. With that said, we've had performance issues with the current front-end and while I'm pretty sure a large part of the slowness I experienced during last year's mystery hunt was from the regression fixed in #211 the existing DataTables + TreeGrid libraries are in part just fundamentally slow. Using the data from last year's Mystery Hunt, the actual table rendering currently takes 1.1s on my machine, of which 600ms is from TreeGrid and the other 500ms come from DataTables, and that renders only 50 puzzles. The React version only takes 460ms to render all 175 puzzles, and while that will go up some once I add the remaining functionality and styling, it should still come in a lot faster. 

Beyond performance, React is now ubiquitous in modern front-end codebases and should be easier to work with, maintain, and extend. It has good tooling and extensive library support. 

This branch currently changes the following:

- Adds a new Django app called `api` with the beginnings of a REST api for smallboard. At the moment this API has two endpoints: one to get info about a hunt and one to get a list of all puzzles of a hunt. It uses Django's sessions authentication but only checks right now that you are logged in.

- Adds a new view, at /hunts/react/:id, to render the all puzzles page with React.

- Adds some React code to call the API endpoints and use the data to render a barebones table. The table supports expand/collapse and filtering in the same ways as existing smallboard, handles the puzzle ordering and the tree construction itself, and renders all puzzles at once (no pagination).

- Adds new dependencies on django rest framework, yarn (for the JS dependencies), babel & webpack (transpilation and bundling), react, and react-table (which I have had to modify/extend to get filtering to work correctly).

- Update the Docker configs to handle JS compilation and reloading.

At this point I think the riskiest/most complicated pieces are done and I feel pretty good that this is going to produce something viable, but there are still a lot of tasks to get it production-ready and at parity with or better than the existing front-end:

- [x] Real-time table updates
    - [x] Poll the puzzles API every 10s, as the current UI does.
    - [x] Make the collapse logic preserve collapse state across updates and not re-expand everything. Current smallboard doesn't do this; it's going to be complicated to get fully correct since we may need to map rows pre-update to rows post-update and that's not always going to be trivial: there may be duplicate rows (puzzles assigned to multiple metas) in different collapsed states, puzzles could be deleted or have their meta relationships change, etc, but we should be able to get the common cases correct at least.
- [ ] Port functionality (UI + API)
    - [ ] Adding a puzzle
    - [ ] Deleting a puzzle
    - [ ] Editing a puzzle
    - [ ] Editing puzzle tags
    - [ ] Meta assignment
    - [ ] Answer submission
    - [ ] Active puzzler display
-  [ ] Improve UI
    - [ ] Style everything
    - [ ] Render the correct table entries: tags, linkify URLs, remove extra debugging columns.
    - [ ] Display some basic stats (number of puzzles visible/total, num solved/pending/stuck/etc)
- [ ] Productionize
    - [ ] Integrate webpack into the deployment pipeline
    - [x] Make the JS `prettier`
    - [ ] Add testing, especially for the API
    - [ ] Update instructions and documentation

For next steps, I'm going to work on the real-time updates and persisting the collapse/expand when they happen, since I think it's the most complicated and will require doing more work with react-table, then some of the UI improvements, then work through porting the other pieces of functionality. At this point a lot of stuff is parallelizable if anyone else is interested in contributing as well. I'm planning on continuing to work on this extensively for the rest of the week and hoping to demo it at our next sync this Sunday.


akirabaruah:
Thanks Ryan - should we track those outstanding tasks you listed in some new Issues to help facilitate splitting up the work?

Since this PR is already quite large, I assume the plan is to first merge the changes in this PR to get the basic structure, then followup with the PRs for the improvements you listed above - does that make sense here?

akirabaruah:
> Thanks Ryan - should we track those outstanding tasks you listed in some new Issues to help facilitate splitting up the work?
> 
> Since this PR is already quite large, I assume the plan is to first merge the changes in this PR to get the basic structure, then followup with the PRs for the improvements you listed above - does that make sense here?

Or is the plan to put all of those changes on this branch/PR? Seems like that might be harder to merge cleanly :open_mouth:

akirabaruah:
@pcvera PTAL too :smile:

hunts/src/hunt-view.js
```diff
@@ -0,0 +1,161 @@
+import React from "react";
```
akirabaruah, line 1: Isn't it a convention in React to name component files using upper camelcase to match their exported component name? e.g. `HuntViewMain.js` in this case

akirabaruah (commented):
Ah gotta love React :heart: LGTM overall with some questions. I'm a React noob, so I'm hoping others like @cs-miller @pcvera @rawxfish can chime in on any details.

A couple other ideas from reading this PR (probably beyond the scope of this PR):
- We should probably add some `prettier` formatting to the pre-commit checks
- Should we use typescript?

cs-miller:
+1 for prettier, +1 for ts

cs-miller:
would suggest using some abstraction over webpack if we can get away with it, not sure if that would be more or less difficult to fit into existing pipeline

hunts/src/puzzle-table.js
```diff
@@ -0,0 +1,120 @@
...
+}) {
+    const [value, setValue] = React.useState(globalFilter);
+
+    return (
+        <span>
+            <input
+                value={value || ""}
+                onChange={e => {
+                    setValue(e.target.value);
+                    setGlobalFilter(e.target.value);
```
pcvera, line 114: If the local value and the value owned by global filter are just being kept in synch here, could we just use the value from `globalFilter` as the source of truth?

hunts/src/puzzle-table.js
```diff
@@ -0,0 +1,120 @@
...
+    const keys = ["values.name", "values.tags", "values.status", "values.answer"];
+    // Need to clone these results because this library is broken as shit
+    return words.reduceRight(
+        (results, word) => matchSorter(results, word, { keys }),
+        rows).map(row => Object.assign({}, row));
+}
+
+textFilterFn.autoRemove = val => !val;
+
+function usePuzzleTable({ columns, data }) {
```
pcvera, line 38: In the context of React, the `use` prefix is associated with hooks.

Dockerfile
```diff
@@ -6,8 +6,14 @@ WORKDIR /usr/src/smallboard
...
     && apk add postgresql-dev gcc python3-dev musl-dev
 
+# install npm & yarn
+RUN apk add --update nodejs yarn
+
 RUN pip install --upgrade pip
 COPY ./requirements.txt .
 RUN pip install -r requirements.txt
 
+COPY ./package.json .
```
cs-miller, line 16: think you will want the lockfile as well

docker-entrypoint.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+yarn run dev &
```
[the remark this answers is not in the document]
cs-miller: i think the idea is that it's hot reloading

hunts/src/puzzle-table.js
```diff
@@ -0,0 +1,120 @@
...
+        return rows;
+    }
+
+    const words = filterValue.split(' ');
+    if (!words) {
+        return rows;
+    }
+
+    const keys = ["values.name", "values.tags", "values.status", "values.answer"];
+    // Need to clone these results because this library is broken as shit
```
cs-miller, line 30: match-sorter or react-table?

hunts/src/puzzle-table.js
```diff
@@ -0,0 +1,120 @@
...
+            data,
+            filterTypes,
+            autoResetExpanded: false,
+            globalFilter: "globalFilter",
+        },
+        useGlobalFilter,
+        useExpanded,
+    );
+
+    React.useMemo(() => toggleAllRowsExpanded(true), [data]);
```
cs-miller, line 69: I think you want an effect with data as the dep

hunts/src/use-global-filter-fixed.js
```diff
@@ -0,0 +1,219 @@
...
+}
+
+function shouldAutoRemoveFilter(autoRemove, value, column) {
+  return autoRemove ? autoRemove(value, column) : typeof value === 'undefined'
+}
+
+// End of utils.js copying
+
+function reducer(state, action, previousState, instance) {
+  if (action.type === actions.resetGlobalFilter) {
```
cs-miller, line 54: would recommend using a switch statement on action.type

rgossiaux: To be clear I didn't write this code and don't intend on modifying it. This file is modified from the react-table library, you can see the changes in the commit log.

hunts/src/puzzle-table.js
```diff
@@ -0,0 +1,120 @@
...
+        prepareRow,
+        toggleAllRowsExpanded,
+        flatRows,
+        preGlobalFilteredRows,
+        setGlobalFilter,
+    } = useTable(
+        {
+            columns,
+            data,
+            filterTypes,
```
pcvera, line 61: Having read through this PR super thoroughly, I'm pretty confident we could cut a lot of complexity by home-rolling our filtering and _probably_ our nesting capabilities.

Fundamentally, our table state is calculable from our data, our filter, and our collapsed nodes, it looks like a lot of this code is fighting `react-table`'s default behaviors, I think we can largely avoid it.

pcvera (changes requested):

docker-entrypoint.sh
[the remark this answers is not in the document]
rgossiaux: Yep, that's right, this starts webpack in the background to recompile after file changes

api/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from . import views
+
+urlpatterns = [
+    path("v1/hunt/<int:pk>", views.HuntAPIView.as_view(), name="hunt_api_view"),
```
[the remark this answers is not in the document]
rgossiaux: Open to suggestions-- I'm not sure what that might look like since I don't think we'd want to use a CURRENT_API_VERSION-type constant since the reason to version the API to begin with is to separate these v1 endpoints from any future v2, and some V1_API="v1" type thing doesn't feel like an improvement

hunts/src/puzzle-table.js
rgossiaux: react-table lol, I don't want to link to the issue since this is a public repo and I think github will notify them, but if you look at issue 2064 in the react-table github the filtering code seems kinda broken

hunts/src/puzzle-table.js
rgossiaux: Possibly, this code is based off some examples from the react-table library and I didn't really look into it closely

api/urls.py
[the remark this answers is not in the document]
akirabaruah: Oh yeah, using a constant won't make as much sense if we want to maintain the various version endpoints. 

Maybe we can use `include()` to nest URLs? (see end of this section: https://docs.djangoproject.com/en/3.1/topics/http/urls/#including-other-urlconfs) That might at least deduplicate the "v1" string for each path by grouping them all together.

tbh, it's probably fine to leave these routes as is, and address organization of the version endpoints once we actually switch versions 😄

hunts/src/puzzle-table.js
rgossiaux: I'm actually not sure all that much of it right now is fighting react-table, I think only a couple lines in this file are doing that right now (but the `-fixed.js` file changes definitely do fall in that bucket). At this point I have little affection for this library but I'm not sure if it's worth writing our own version.

rgossiaux:
Thanks for the comments, I'm glad that you guys are interested and invested 😄 I'll try to incorporate some of the feedback.

Re: prettier, it's on the list! There are some other QOL changes that you guys have noticed too which I think are also important, eg I started writing this with class-based components since that's what I'm familiar with but react-table forced me to use hooks so that's why both are there; probably the class one should use hooks for consistency, although i do personally find them much less clear. Similarly while I thought of using typescript I haven't actually used it before and wanted to focus on productivity vs learning a new language (and I was also a little worried that it might make it harder for other people to contribute or fix things).

cs-miller:
im with you on typescript raising the difficulty level, so probably off the table. side comment, if you want to make these changes to the react-table library without this "-fixed" business, you could use [patch-package](https://www.npmjs.com/package/patch-package)

hunts/src/puzzle-table.js
cs-miller: it's funny to me how much trouble this lib seems to be giving us, i see the guy shilling it on twitter non-stop 😆

rgossiaux:
> side comment, if you want to make these changes to the react-table library without this "-fixed" business, you could use [patch-package](https://www.npmjs.com/package/patch-package)

Wow, this looks fantastic, thanks! Will definitely look into it.

hunts/src/hunt-view.js
rgossiaux: yeah, I've always disliked camelcase in file names but i'll change it

akirabaruah:
Avoiding camel case names is fine too as long as we've got a defined style
IMO. Should we just pick someone's style guide (maybe airbnb or standard
js?) and plug it into prettier?

cs-miller:
That’s more of an eslint thing

cs-miller:
Which we definitely should have (eslint)

rgossiaux:
Ok, I tried to address most of the comments above, feel free to take a look. 

I did switch to `patch-package` for patching react-table though it was surprisingly nontrivial-- since we need to patch the dist/ files from the package, i needed to rebuild react-table first to regenerate them, and you wouldn't believe how many dependencies it has (well, you probably would, because that's what the npm world looks like these days). 

Very amusingly I needed to `patch-package` `patch-package` itself because it doesn't use a big enough buffer to handle the patch.

cs-miller:
are you saying it takes running yarn install 2x to have patch package work? i just ran npm install and had it work no problem afaict.

hunts/src/hunt-view.js
```diff
@@ -1,161 +1,141 @@
...
-            return -1;
-        } else if (a.feeders.length > 0 && b.feeders.length == 0) {
-            return 1;
-        }
-        // Newer puzzles before old ones
-        // TODO: once creation times are added to puzzles, use those instead
-        return b.id - a.id;
-    }
+export const HuntViewMain = (props) => {
+  const [huntData, setHuntData] = React.useState(props);
```
cs-miller: should the initial state be props here? seems like it should just be nothing.

rgossiaux:
> are you saying it takes running yarn install 2x to have patch package work? i just ran npm install and had it work no problem afaict.

You're talking about the Dockerfile? I structured it that way so that adding patches doesn't require re-installing all the npm dependencies from scratch. Docker images are built in cached layers, so if we first copied both package.json and the patches and then ran `yarn install` in one go, changes to the patches would invalidate the cache for all the node_modules. As it is the second `yarn install` doesn't do anything besides applying the patches and takes less than a second on my machine. With that said I'm not sure if this is better overall or not but that was the idea at least.

cs-miller:
ok yeah figured it was just for the layers

hunts/src/hunt-view.js
```diff
@@ -1,161 +1,141 @@
...
+      .then((puzzleData) => {
+        setPuzzleData(processPuzzleData(puzzleData));
+      });
+  }, [props.huntId]);
+
+  if (huntData) {
+    return (
+      <div>
+        <h1>{huntData.name} - All Puzzles</h1>
+        <PuzzleTable columns={TABLE_COLUMNS} data={puzzleData} />
```
cs-miller: could potentially pass a "loading" prop to PuzzleTable while puzzleData is loading

hunts/src/puzzle-table.js
```diff
@@ -1,120 +1,111 @@
...
 
-function usePuzzleTable({ columns, data }) {
-    const filterTypes = React.useMemo(
-        () => ({
-            globalFilter: textFilterFn,
-        }),
-        []
-    );
+export function PuzzleTable({ columns, data }) {
+  const filterTypes = React.useMemo(
```
cs-miller, line 26: i think this could be lifted out of the component?

rgossiaux:
Ok, just added in the live updating logic and got the table to preserve state across updates (our default-expanded behavior meant I had to make another patch to `react-table`, which defaults to collapsed). I think that should be just about all of the hairy table things done, so going forward should now just be writing standard UI/API code and no more source diving into react-table internals.

I'd like to merge this into master soon since the PR is already kind of large and so people can work on it without having to merge into my branch. We can then address any follow-ups & future work separately. Let me know what you think; @pcvera since you requested changes I'll need you to approve to merge.

pcvera (approved):

PR #238, opened 2020-12-02
iss #197: make homepage fancier

rawxfish:
Looks like this now: 

image


rgossiaux (approved):
Looks good to me, thanks for tackling this! 

Is having the google button a requirement? It's kind of ugly, heh.

rawxfish:
> Is having the google button a requirement? It's kind of ugly, heh.

No. I was just trying to think up ways to please the overlords lol. We can do a style pass later

akirabaruah:
+1 thanks and LGTM 😄  another small thought: what do you folks think of moving this template and route from `/accounts/login` to simply `/` or `/index` since it's now more of a bona fide home page?

PR #242, opened 2020-12-03
Fixes for heroku instance

akirabaruah:
Required for the hunt slug URL change in #201.


rawxfish (approved):

akirabaruah:
Wow that was fast 😮  I'm not yet convinced that this is the full fix as I'm unable to repro the issue locally. We'll need this change regardless.

rawxfish:
I swear I do get work done at work.

PR #243, opened 2020-12-04
move react to externs to reduce our package size

pcvera:
Moves React and ReacDOM to `external`, dropping the size of the main.js we serve by 80%, pushing the size load to CDNs
Introduces a `base-react.html` so we can separate cdn dependencies across the React and jQuery implementations, while they have to live simultaneously.
Adds `water.css` to give us some default styling for no changes to our codebase:
image


rgossiaux:
thanks!! very glad that the table looks better (though does this break the top navigation bar? looks like it a bit in the screenshot)

personally at this point I'm not big on 3rd party CDN libraries any more since the cache benefits (which have always been limited anyway) are disappearing: https://chromestatus.com/feature/5730772021411840 and there are some downsides: https://shkspr.mobi/blog/2020/10/please-stop-using-cdns-for-external-javascript-libraries/ (and more to the point: you can't run the site locally when offline + we're stuck with a minified production bundle when developing). curious what you think about that

cs-miller:
would prefer having dev build in dev if we can

pcvera:
Regarding the top nav, it's actually broken by the removal of bootstrap rather than addition of water, we can keep it, but I didn't think we would ultimately, so I was cool with it breaking because it was an honest break, preferring it over keeping bootstrap to disguise our true state.

Regarding CDN vs building in,  the primary benefit I was looking for was minimizing the load on the smallboard server but I hadn't considered dev mode and that is a pretty compelling problem to me.

rgossiaux:
I wouldn't worry about the server load, other ways to deal with that: we have [whitenoise](http://whitenoise.evans.io/en/stable/) installed already which should be sufficient, and smallboard.app is also behind cloudflare so we can leverage their CDN if we need to.

cs-miller:
if we want this cdn business its very doable in webpack for the "prod" build using HTMLWebpackPlugin, although i'm a little unclear about how we are interacting with django atm.

PR #244, opened 2020-12-04
add prettier to our codebase and run it over the existing javascript

pcvera:
no logic changes


rgossiaux (approved):
Thanks! What do you think about adding to our github actions config?

akirabaruah:
> Thanks! What do you think about adding to our github actions config?

Actually, WDYT about adding a pre-commit check for prettier (similar to black for python)? Then you won't have to update the github actions since they already run the pre-commit checks.

pcvera:
I went ahead and added prettier to the precommit

akirabaruah:
Seemed like pre-commit's yaml parser was grumpy, so I fixed the indentation. Thanks @pcvera!

(looks like we might have to rebase this branch in order to merge)

yarn.lock
```diff
@@ -2350,6 +2350,8 @@ pkg-dir@^4.1.0, pkg-dir@^4.2.0:
   dependencies:
     find-up "^4.0.0"
 
+<<<<<<< HEAD
+=======
```
erwa, line 2354: Looks like this was accidentally committed

yarn.lock
```diff
@@ -2360,6 +2362,7 @@ postinstall-postinstall@^2.1.0:
   resolved "https://registry.yarnpkg.com/postinstall-postinstall/-/postinstall-postinstall-2.1.0.tgz#4f7f77441ef539d1512c40bd04c71b06a4704ca3"
   integrity sha512-7hQX6ZlZXIoRiWNrbMQaLzUUfH+sSx39u8EJ9HYuDc1kLo9IXKWjM5RSquZN1ad5GnH8CGFM78fsAAQi3OKEEQ==
 
+>>>>>>> rgossiaux/react
```
erwa, line 2365: same

erwa (commented):
Looks like some unresolved conflicts were merged

yarn.lock
erwa: I'm surprised a precommit check didn't catch this

akirabaruah: Yeah, we have a pre-commit check that runs during `git merge` so the only time it would have run was during the merge commit above. @pcvera is it possible you don't have `pre-commit` hooks installed in your local git repo?

In any case, I'll put up a PR to enable merge conflict lint checking always so we can catch these in Github :+1: Thanks for pointing it out, @erwa!

akirabaruah: See #247 for the fix

akirabaruah:
Addressed in #247 :+1:

PR #245, opened 2020-12-04
Populate slug field for existing hunts

akirabaruah:
Adds a data migration for existing hunts that creates a slug based on
the hunt's primary key, which is already guaranteed to be unique.

This migration ensures that existing dbs aren't broken by #201.


hunts/migrations/0003_populate_hunt_slug.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.1.4 on 2020-12-04 07:47
+
+from django.db import migrations
+from django.template.defaultfilters import slugify
+
+
+def slugify_existing_hunts(apps, schema_editor):
+    Hunt = apps.get_model("hunts", "Hunt")
+    for hunt in Hunt.objects.all():
+        hunt.slug = slugify(hunt.pk)
```
rgossiaux, line 10: What do you think about using `hunt.name` here like I think you're already doing for the production code? I think I might have told you to use the ID before but your approach of calling this `slugify` function on the hunt name seems much better.

Pretty minor but we might want to check if the slug already exists before setting it, so that if this migration ever gets rerun for some reason it won't blow away existing user-chosen slugs.

rgossiaux:
Thank you!!

hunts/migrations/0003_populate_hunt_slug.py
akirabaruah: Both suggestions SGTM :+1:

akirabaruah: Actually, at that point, we're just using the exact same logic as the `Hunt.save()` override, so to keep things DRY, we can just re-save all existing hunts.

rgossiaux (approved):
Makes sense, thanks!

erwa (approved):

PR #246, opened 2020-12-04
Rename pre-commit workflow to pre-submit

akirabaruah:
~~This relaxes the requirements for merging PRs to other PRs or non-master branches.

@pcvera since you're authoring PRs targeting @rgossiaux's PR for the react rewrite, I figure we shouldn't block the changes on presubmit checks until it's time to merge the whole thing into master (a.k.a. the main branch).~~

@erwa pointed out that we can still leave these presubmits to run on any PR without blocking merging as long as we don't merge into the master (protected) branch. In that case, this PR just renames the workflow as a cleanup.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -2,6 +2,7 @@ name: pre-commit
 
 on:
   pull_request:
+    branches: [master]
   push:
     branches: [master]
 
@@ -18,7 +19,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.8.x'
+          python-version: "3.8.x"
 
       - name: Install python deps
         run: |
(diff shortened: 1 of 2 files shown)


erwa:
I think it's still helpful to run precommit checks when merging to non-master branches. Note, since other branches are not protected, failing checks does not block merging to non-master branches. E.g., in #244, you can still merge even though checks have failed:
image

akirabaruah:
Oh good to know - thanks for pointing it out @erwa - let's scrap this PR then 😄

akirabaruah:
Oh wait, I guess the rename is still helpful, so I can just revert the pull_request change - wdyt?

erwa (approved):
Sure, I'm fine with renaming it to pre-submit. I don't feel strongly about this either way.

akirabaruah:
Yeah same :+1: figured it was one of those "might as well" situations :smile:

PR #247, opened 2020-12-06
Clean up pre-commit checks

akirabaruah:
- always run check-merge-conflict
- remove requirements.txt fixer (since we don't use that file anymore)


erwa (approved):
Thanks for adding this!

Issue #251, opened 2020-12-07
Rewrite main hunt view in React 


rgossiaux:
Following on from #237 

You can view the progress locally by running `docker-compose up` and going to /hunts/react/[slug]. 

Copying over the task list from the old PR:

- [x] Real-time table updates
    - [x] Poll the puzzles API every 10s, as the current UI does.
    - [x] Make the collapse logic preserve collapse state across updates and not re-expand everything. Current smallboard doesn't do this; it's going to be complicated to get fully correct since we may need to map rows pre-update to rows post-update and that's not always going to be trivial: there may be duplicate rows (puzzles assigned to multiple metas) in different collapsed states, puzzles could be deleted or have their meta relationships change, etc, but we should be able to get the common cases correct at least.
- [x] Port functionality (UI + API)
    - [x] Adding a puzzle
    - [x] Deleting a puzzle
    - [x] Editing a puzzle
    - [x] Editing puzzle tags
    - [x] Meta assignment (@rgossiaux )
    - [x] Answer submission
    - [ ] Active puzzler display (probably deprecated)
-  [ ] Improve UI
    - [x] Style everything
    - [x] Render the correct table entries: tags, linkify URLs, remove extra debugging columns.
    - [ ] Display some basic stats (number of puzzles visible/total, num solved/pending/stuck/etc)
- [ ] Productionize
    - [x] Integrate webpack into the deployment pipeline
    - [x] Make the JS `prettier`
    - [x] Add testing, especially for the API
    - [ ] Update instructions and documentation

rgossiaux:
Going to tackle some of the "improve UI" pieces next (especially rendering the right things in the table)

rgossiaux:
Update: I implemented a first version of the delete puzzle modal/API locally (depends on #253) but the data flow is already kinda ugly. I'm refactoring it to use redux which should make the code a lot nicer.

I'm starting work on Monday so I'm going to try to get as much as I can done before then.

rgossiaux:
I've linked a lot of things here so for convenience the current set of PRs is #253 => #256 => #257 => #258 .

It turns out my equipment for work won't arrive for at least a week so in the meantime I still have a lot of time to work on this. Hoping that by our meeting this Sunday we will almost be at the point where we could switch over. 

[...]

PR #253, opened 2020-12-07
[React] render all the table elements

rgossiaux:
For #251. Using react-bootstrap to mimic the existing UI. I made what I thought were a few improvements: bringing the tag delete button inside the tags (looks way better), changing the status to a dropdown, pulling puzzle links into a separate column (I found it confusing last year to have the names link to the puzzles; here meta names can be clicked for expand/collapse instead), and some minor other changes.

Here's what it looks like now: 
image

None of the buttons do anything yet, the next step is to hook them all up to API calls.


hunts/src/HuntViewMain.js
```diff
@@ -2,53 +2,130 @@ import React from "react";
...
+        <span className="text-monospace">{value}</span>
+      ) : (
+        <Button variant="outline-primary">Submit Answer</Button>
+      ),
   },
   {
     Header: "Status",
     accessor: "status",
+    Cell: ({ row, value }) =>
+      ["SOLVING", "STUCK", "EXTRACTION"].includes(value) ? (
```
erwa, line 62: Something to consider (not sure if worthwhile or practical) -- is it possible to define these possible Status states in some common enum file and refer to the enum constants here and elsewhere instead of hardcoding strings? If we used enum constants, I wonder if we could catch typos via some static typechecking (or do we need TypeScript for that)?

hunts/src/HuntViewMain.js
```diff
@@ -2,53 +2,130 @@ import React from "react";
...
+        <DropdownButton
+          id={`status-dropdown-${row.id}`}
+          title={value}
+          variant="outline-primary"
+        >
+          <Dropdown.Item>SOLVING</Dropdown.Item>
+          <Dropdown.Item>STUCK</Dropdown.Item>
+          <Dropdown.Item>EXTRACTION</Dropdown.Item>
```
erwa, line 70: Should we generate this using some sort of map() or for loop?

erwa (approved):
I'm not familiar with React or frontend frameworks in general, but these UI changes look fine to me. I agree that the tags look better with the X button inside them.

hunts/src/HuntViewMain.js
rgossiaux: Yeah this is definitely a good idea, I'll pull some constants out to another file.

I'm supportive of migrating this to typescript but stuck with es6 initially since I haven't used typescript before, and I wanted to finish this rewrite as soon as I could since it's going to clash with other changes we want to make before mystery hunt. But I do want to make the switch maybe after mystery hunt this year.

PR #254, opened 2020-12-08
Move sheet update logic from views to signal hooks

rgossiaux:
Simpler to just handle this in one place here. I tested adding/removing
metas with "Assign Metas" and with tags as well as deleting puzzles.


erwa (approved):
Nice! This definitely seems cleaner than the scattered logic we had before. Good use of hooks.

PR #255, opened 2020-12-11
[React] Add redux, puzzle deletion UI/API

rgossiaux:
Exactly the same interface as the current site. For #251, depends on #253 

Originally I wrote this UI without redux (mostly cuz I've never used it) but triggering things from inside the individual table cells (like opening a modal) was already getting complicated and there was no good way to make the UI reflect the changes instantly without waiting for a full refresh. So I bit the bullet and worked through some redux docs... I'm sure there's some unidiomatic or outright bad stuff in here.

There's a lotta boilerplate/scaffolding but hopefully now the other actions can just follow these patterns.


rgossiaux:
I kinda fucked up the git history here, moving to #256

PR #256, opened 2020-12-11
[React] Add redux, puzzle deletion UI/API

rgossiaux:
Exactly the same interface as the current site. For #251, depends on #253 

Originally I wrote this UI without redux (mostly cuz I've never used it) but triggering things from inside the individual table cells (like opening a modal) was already getting complicated and there was no good way to make the UI reflect the changes instantly without waiting for a full refresh. So I bit the bullet and worked through some redux docs... I'm sure there's some unidiomatic or outright bad stuff in here.

There's a lotta boilerplate/scaffolding but hopefully now the other actions can just follow these patterns.

### DIFF
--- a/hunts/src/index.js
+++ b/hunts/src/index.js
@@ -1,1 +1,2 @@
+import "regenerator-runtime/runtime.js";
 import App from "./App";
--- a/hunts/src/puzzlesSlice.js
+++ b/hunts/src/puzzlesSlice.js
@@ -1,0 +1,102 @@
+import {
+  createSelector,
+  createSlice,
+  createAsyncThunk,
+  createEntityAdapter,
+} from "@reduxjs/toolkit";
+import api from "./api";
+
+export const deletePuzzle = createAsyncThunk(
+  "puzzles/deletePuzzle",
+  async ({ huntId, id }) => {
+    const response = await api.deletePuzzle(huntId, id);
+    return id;
+  }
+);
+
+export const fetchPuzzles = createAsyncThunk(
+  "puzzles/fetchPuzzles",
+  async (huntId) => {
+    const response = await api.getPuzzles(huntId);
+    return response;
+  }
+);
+
+function puzzleComparator(a, b) {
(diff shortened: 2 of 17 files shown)


pcvera (approved):
Surprised you've never done redux before, this code all looks really solid.

Just a small comment, in a lot of cases thunks can be replaced with a normal async function followed by a dispatch and in cases where they're more complicated than that they're usually better served by other libraries.

hunts/src/puzzlesSlice.js
```diff
@@ -0,0 +1,102 @@
...
+);
+
+function puzzleComparator(a, b) {
+  // Solved puzzles should appear below unsolved ones
+  if (a.status == "SOLVED" && b.status != "SOLVED") {
+    return 1;
+  } else if (b.status == "SOLVED" && a.status != "SOLVED") {
+    return -1;
+  }
+  // Feeders before metas
```
asdfryan, line 32: Should feeders really come before metas? It seems more reasonable for feeders to be listed after the metas:

meta
   feeder1
   feeder2

etc.

asdfryan (approved):
LGTM

hunts/src/puzzlesSlice.js
rgossiaux: This is for sorting puzzles at the same depth of the tree, so if `feeder1` and `feeder2` are assigned to `meta`, they'll appear below `meta` as one of the children. I think if `feeder1` and `feeder2` are the same depth as `meta` then the mostly likely thing is that they're new puzzles that haven't been assigned a meta or whose meta hasn't been unlocked yet, and in that case it makes sense to me to keep them above the meta block so that they're more visible (for example imagine if the subtree of `meta` was 100 puzzles).

hunts/src/index.js
```diff
@@ -1 +1,2 @@
+import "regenerator-runtime/runtime.js";
```
erwa, line 1: Is this needed? Does React/Redux require generator functions, or are you using any?

Issue #196, opened 2020-11-24
Support multiple answers per puzzle


erwa:
For MITMH 2020, there were some puzzles that had multiple answers (e.g.: the [Safari Adventure puzzles](https://www.mit.edu/~puzzle/2020/land/safari/)). It would be nice to support this on Smallboard. That said, multiple answers per puzzle probably won't come up in many hunts, so this is not that high priority, I'd say.

asdfryan:
Totally forgot about this -- how'd we deal with this last year?

rawxfish:
We just concatenated all of the answers and manually separated them later

rgossiaux:
We have support for this now as of #275 although I think we still need to iterate on the UI (tracked in #280)

PR #257, opened 2020-12-12
[React] Add API/UI for editing puzzles and changing puzzle status

rgossiaux:
They're implemented separately in the commit history to view them separately. Once the API to edit a puzzle is in place, changing puzzle status (only between solving/stuck/extraction) is nearly free fortunately.

For #251; depends on #256 and transitively on #253.

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -15,6 +15,30 @@ class PuzzleSerializer(serializers.ModelSerializer):
     def get_tags(self, obj):
         return [{"name": tag.name, "color": tag.color} for tag in obj.tags.all()]
 
+    def validate(self, data, *args, **kwargs):
+        # The django rest framework validation is super clunky for validating
+        # partial updates; we need to merge new values with existing ones
+        # ourselves annoyingly.
+        def get_merged(attr):
+            if attr in data:
+                return data[attr]
+            if self.instance and hasattr(self.instance, attr):
+                return getattr(self.instance, attr)
+            return None
+
+        # Feeders can't have feeders
+        if not get_merged("is_meta") and len(self.instance.feeders.all()) > 0:
+            raise serializers.ValidationError(
+                "Puzzle must be a meta to have puzzles assigned."
+            )
+        # Answers mean solves
+        if get_merged("status") == "SOLVED" and not get_merged("answer"):
+            raise serializers.ValidationError("Solved puzzles must have answers.")
+        if get_merged("answer") and get_merged("status") != "SOLVED":
+            raise serializers.ValidationError("Puzzles with answers are solved.")
+
+        return data
+
     class Meta:
         model = Puzzle
         fields = (
@@ -30,4 +54,14 @@ class PuzzleSerializer(serializers.ModelSerializer):
             "metas",
             "feeders",
             "is_meta",
+        )
+
+        read_only_fields = (
+            "id",
+            "hunt_id",
+            "sheet",
+            "answer",
+            "tags",
+            "metas",
+            "feeders",
         )
--- a/api/views.py
+++ b/api/views.py
@@ -50,3 +50,16 @@ class PuzzleViewSet(viewsets.ModelViewSet):
             puzzle.delete()
 
         return Response({})
+
+    def partial_update(self, request, pk=None, **kwargs):
+        puzzle = None
+        with transaction.atomic():
+            puzzle = self.get_object()
+            serializer = self.get_serializer(puzzle, data=request.data, partial=True)
+            serializer.is_valid(raise_exception=True)
+            data = serializer.validated_data
+            puzzle.update_metadata(
+                new_name=data["name"], new_url=data["url"], new_is_meta=data["is_meta"]
+            )
+
+        return Response(PuzzleSerializer(puzzle).data)
--- a/api/views.py
+++ b/api/views.py
@@ -59,7 +59,12 @@ class PuzzleViewSet(viewsets.ModelViewSet):
             serializer.is_valid(raise_exception=True)
             data = serializer.validated_data
             puzzle.update_metadata(
-                new_name=data["name"], new_url=data["url"], new_is_meta=data["is_meta"]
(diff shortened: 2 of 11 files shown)


pcvera (approved):
JS looks good

api/views.py
```diff
@@ -50,3 +50,21 @@ def destroy(self, request, pk=None, **kwargs):
...
         return Response({})
+
+    def partial_update(self, request, pk=None, **kwargs):
+        puzzle = None
+        with transaction.atomic():
+            puzzle = self.get_object()
+            serializer = self.get_serializer(puzzle, data=request.data, partial=True)
+            serializer.is_valid(raise_exception=True)
+            data = serializer.validated_data
+            puzzle.update_metadata(
```
asdfryan, line 61: I think `update_metadata()` already calls save()?

rgossiaux: Yeah, but the call on line 68 is for saving the change on line 67 when we're changing the puzzle status (stuck/extraction/etc), which is not handled by update_metadata. You could probably refactor this so that when we're updating both eg the puzzle name and the status at the same time (not supported by the UI currently but supported by the API) we only make one call to `.save()` instead of two but for now I'm not worried about it

api/serializers.py
```diff
@@ -15,6 +15,30 @@ class PuzzleSerializer(serializers.ModelSerializer):
...
+
+        # Feeders can't have feeders
+        if not get_merged("is_meta") and len(self.instance.feeders.all()) > 0:
+            raise serializers.ValidationError(
+                "Puzzle must be a meta to have puzzles assigned."
+            )
+        # Answers mean solves
+        if get_merged("status") == "SOLVED" and not get_merged("answer"):
+            raise serializers.ValidationError("Solved puzzles must have answers.")
+        if get_merged("answer") and get_merged("status") != "SOLVED":
```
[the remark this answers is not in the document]
asdfryan: Maybe add a TODO comment (with a linked issue) so we remember to deal with this when we support multiple answers.

asdfryan (approved):

api/serializers.py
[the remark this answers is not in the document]
rgossiaux: The puzzle model has both a denormalized answer field (`puzzle.answer`) and a relationship with the answer model (accessible as `puzzle.guesses`). The answer queue stuff is only relevant to the answer model. Right now puzzles can have multiple guesses but only one answer, but yeah if we do change that for #196 (eg by just removing the answer field and looking at the answer model) then we'll have to update this code (as well as other places that use `puzzle.answer`, like the front-end).

api/views.py
erwa: Hmm... it seems a little strange to have two save paths: puzzzle.update_metadata() and puzzle.save(). Any chance these can be combined? Also, I see that update_metadata() calls update_sheets() on commit. When status changes (e.g.: we have a new confirmed answer for a meta) and puzzle.save() is called, is update_sheets() also called?

hunts/src/puzzlesSlice.js
```diff
@@ -55,6 +63,12 @@ export const puzzlesSlice = createSlice({
     [fetchPuzzles.fulfilled]: (state, action) => {
       puzzlesAdapter.setAll(state, action.payload);
     },
+    [updatePuzzle.fulfilled]: (state, action) => {
+      puzzlesAdapter.updateOne(state, {
+        id: action.payload.id,
+        changes: { ...action.payload },
+      });
+    },
```
erwa, line 71: Is there any indication for .rejected? We can add it later, though it'd be helpful to add a TODO as a reminder.

api/serializers.py
```diff
@@ -31,3 +55,13 @@ class Meta:
...
+
+        read_only_fields = (
+            "id",
+            "hunt_id",
+            "sheet",
+            "answer",
+            "tags",
+            "metas",
+            "feeders",
+        )
```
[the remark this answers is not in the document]
rgossiaux: I'm not sure what the precise effects are of setting this (whether it will complain at deserialization or when saving the object or what) but yes the idea is to prevent update puzzle API calls from changing them. I'll loosen this as necessary when it comes up (though for tags I created a separate /tags endpoint for adding/removing them).

api/serializers.py
```diff
@@ -15,6 +15,30 @@ class PuzzleSerializer(serializers.ModelSerializer):
...
 
+    def validate(self, data, *args, **kwargs):
+        # The django rest framework validation is super clunky for validating
+        # partial updates; we need to merge new values with existing ones
+        # ourselves annoyingly.
+        def get_merged(attr):
+            if attr in data:
+                return data[attr]
+            if self.instance and hasattr(self.instance, attr):
+                return getattr(self.instance, attr)
```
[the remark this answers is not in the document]
rgossiaux: The first is checking `data` (the new values passed by the client), the second is checking `self.instance` (the existing object)--this method is to get what the value of the field will be after the update, to use in the validation checks below.

hunts/src/puzzlesSlice.js
rgossiaux: Error handling is done elsewhere: https://github.com/cardinalitypuzzles/smallboard/blob/master/hunts/src/alertSlice.js#L29-L34 for displaying error alerts (similar to the non-react UI)

api/views.py
rgossiaux: Yeah, I agree. For the moment I just wanted to focus on porting over the front-end and the existing structure as much as possible without also rewriting the model code, but I'd like to make this part better too. Ideally I think we get rid of puzzle.update_metadata and move some of the logic elsewhere (inside puzzle.save(), or pre/post_save hooks, or checks in the view methods).

As for status changes, by themselves right now they don't affect the sheets (eg a puzzle moving from SOLVING -> STUCK). Adding an answer isn't supported yet but when it is I'll make sure it updates the sheets.

erwa: I filed #271 for updating sheets when updating answers so we don't forget

PR #258, opened 2020-12-15
[React] add Add Puzzle API/UI

rgossiaux:
Reasonably straightforward. I also included some small UI cleanups and polish (autofocus, front-end form validation)

### DIFF
--- a/hunts/src/api.js
+++ b/hunts/src/api.js
@@ -3,9 +3,27 @@ import Cookies from "js-cookie";
 function getPuzzles(huntId) {
   const puzzlesApiUrl = `/api/v1/hunt/${huntId}/puzzles`;
   return fetch(puzzlesApiUrl).then((response) => {
-    if (response.status > 400) {
+    if (response.status >= 400) {
       // TODO: error handling
-      console.error("Get puzzles API failure", response);
+      throw new Error("Get puzzles API failure");
+    }
+    return response.json();
+  });
+}
+
+function addPuzzle(huntId, data) {
+  const puzzlesApiUrl = `/api/v1/hunt/${huntId}/puzzles`;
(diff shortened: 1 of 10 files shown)


rgossiaux:
(for #251 )

hunts/src/api.js
```diff
@@ -3,9 +3,27 @@ import Cookies from "js-cookie";
...
+  });
+}
+
+function addPuzzle(huntId, data) {
+  const puzzlesApiUrl = `/api/v1/hunt/${huntId}/puzzles`;
+  return fetch(puzzlesApiUrl, {
+    method: "POST",
+    headers: {
+      "X-CSRFToken": Cookies.get("csrftoken"),
+      "Content-Type": "application/json",
```
akirabaruah, line 20: Seems like we use the same headers for various requests throughout these React forms. Is there a best practice to avoid repeating these? Or is it more conventional to explicitly repeat them?

akirabaruah (approved):
LGTM

asdfryan (approved):

hunts/src/api.js
```diff
@@ -17,9 +35,9 @@ function deletePuzzle(huntId, puzzleId) {
...
     headers: { "X-CSRFToken": Cookies.get("csrftoken") },
   }).then((response) => {
-    if (response.status > 400) {
+    if (response.status >= 400) {
       // TODO: error handling
-      console.error("Delete puzzle API failure", response);
+      throw new Error("Delete puzzle API failure");
     }
     return response.json();
   });
```
erwa, line 43: Should we .catch these errors and display an error message at the top of the page? Right now, I think if an error happens, the script will just terminate with an error message in the console.

PR #259, opened 2020-12-16
[React] Error handling and webpack improvements

rgossiaux:
The two commits in this branch aren't really related to each other except that they're things I'm doing to make this more production-ready. Needed to stop running `watch` for the production webpack config, and also added some pieces for error handling, so that errors now generally produce sensible messages.


rgossiaux:
For linking purposes, this is for #251

pcvera (approved):
JS looks good to me

PR #261, opened 2020-12-17
Move tags off of taggit and associate puzzle tags with hunts

rgossiaux:
This is needed to sandbox hunts from each other. Before, all the tags
from hunt 1 would appear in the add tags form for hunt 2, and I think
there would be bad interactions if we ever had eg two instances of the
same hunt on the site or a meta name in one hunt that was also a tag
name in a second hunt.


rgossiaux:
There's a follow-up needed from this to actually remove taggit from the project; right now it's still needed for old migrations but if we merge this we can eventually squash them to remove the dependency.

rgossiaux:
Going to copy out one of the commit messages from this PR: 

```
We didn't really need any of the functionality that taggit provided and
it was pretty hard to work with: trying to write the migration to add
hunt ids to tags was causing me a lot of pain and I've also had
problems trying to use it with django rest framework.

The change in the models themselves is fairly minimal but the migration
to switch from taggit to this new m2m was kind of complicated. I used a
couple SQL queries to save the existing tag relationships to a temp
table and then to dump that table into the m2m relationship that django
created. Then in python I assign hunt_ids to tags based on the puzzles
they're associated with and create new tags when a tag is used in
multiple hunts. The end result is that the transition should preserve
everyone's local data.
```

rawxfish (approved):

Issue #186, opened 2020-11-09
Automatic Discord channel creation (text + voice)


akirabaruah:
Features:
- voice chat
- text chat (similar to existing Slack integration)

rawxfish:
Current plan:

- Ask everyone to switch notifications for the server to "Only @mentions" 
- New puzzle should create text channel under "puzzle" category. The new channel will be announced in an announcements, maybe with @everyone (or we can make this opt-in with a role or sth) 
- Solving puzzle should move text channel from "puzzle" category to "archive" category
- Voice channels can be flexibly created or renamed, and not linked with smallboard. The best practice should be for people to rename VCs to name of puzzles worked on

It might also be nice for smallboard to name text channels with emojis to indicate whether they're meta / high priority / low priority.

rgossiaux:
Just curious, why can't smallboard create audio channels as well?

rawxfish:
We could do that. I was worried it'd be cluttered if we accumulate a lot of open puzzles?

rawxfish:
As a side note, can we name our discord bot Cardi T

akirabaruah:
> We could do that. I was worried it'd be cluttered if we accumulate a lot of open puzzles?

I guess we can collapse the "Voice Channels" tab if needed? Seems like it's the same situation as with text channels? (maybe this was discussed in the last face-to-face sync) Ideally, we'd have some way to group the text and voice channels for a particular puzzle together, but right now, it seems like Discord keeps text and voice separate.

rgossiaux:
Yeah it might be nice for people to not have to worry about creating them, I guess it's going to be more technically complex though. We'd need to delete them after the puzzles got solved, but maybe only when they're empty (it'd be super rude if you just finished a puzzle and got booted out of the voice channel before you could celebrate/discuss with the other solvers), so maybe a cron job (or similar, I think Heroku has a scheduler product) would clean them up periodically, or if that's too much work someone running a script every once in a while

rawxfish:
Is it thaaaaat rude? We could move them to a general "water cooler" voice channel. Might be sorta hype for the people in water cooler since they'd know a puzzle just got solved.

asdfryan:
+1 to also opening a voice channel and periodic clean-up cron (1x an hour should be sufficient IMO). The cron job can take care of both voice + text channels IMO

In the interest of getting this done, can we split up this issue into a couple smaller ones? 
[...]

PR #263, opened 2020-12-18
Start Discord integration

akirabaruah:
@asdfryan here's what @cs-miller and I got working during our last pairing session for #186:

Adds logic to create a new Discord text channel automatically upon
Puzzle creation. The created Discord channel ID gets stored in the
Puzzle model as a discord_channel_id field.

There's still work to:

- [x] add new text channels to a folder/category (this is actually easy to do, but the hard part is managing multiple folders since there's a limit of 50 channels per folder)


rgossiaux:
Nice, glad you're working on this 😄

What do you think about creating a separate model for channels and linking that to puzzles (vs embedding the info in the puzzle model)? I haven't really thought it through so I don't have much opinion one way or another. 

One thing I personally would like is if we can set up some good abstractions around chat so that if/when we switch chat services (as we've already done once lol) we don't have to do everything from scratch all over again. I don't really know what that looks like (and maybe our requirements are too tightly coupled to the specific chat platform we're using) but it might be worth keeping in mind.

hunts/views.py
```diff
@@ -172,9 +173,21 @@ def post(self, request, hunt_slug):
...
+
             try:
                 puzzle = Puzzle.objects.create(
-                    name=name, url=puzzle_url, hunt=hunt, sheet=sheet, is_meta=is_meta
+                    name=name,
+                    url=puzzle_url,
+                    hunt=hunt,
+                    sheet=sheet,
+                    is_meta=is_meta,
+                    discord_channel_id=discord_channel_id,
```
asdfryan, line 190: Do voice and text channels share the same id?

asdfryan (commented):
> Nice, glad you're working on this 😄
> 
> What do you think about creating a separate model for channels and linking that to puzzles (vs embedding the info in the puzzle model)? I haven't really thought it through so I don't have much opinion one way or another.
> 
> One thing I personally would like is if we can set up some good abstractions around chat so that if/when we switch chat services (as we've already done once lol) we don't have to do everything from scratch all over again. I don't really know what that looks like (and maybe our requirements are too tightly coupled to the specific chat platform we're using) but it might be worth keeping in mind.

+1 to this

The channel abstraction could have providers (Zoom/Slack/etc.), channel type (voice, text), and name (slugified, or not) as fields and there could be helper methods to instantiate the channels depending on provider and type.

discord/discord_client.py
```diff
@@ -0,0 +1,72 @@
...
+            and settings.DISCORD_GUILD_ID is not None
+        )
+
+    def __init__(self, api_token, guild_id):
+        if self.__class__.__instance:
+            raise InstantiationError("Singleton object already created.")
+        self._client = APIClient(api_token)
+        self._guild_id = guild_id
+
+    def create_channel(self, name, chan_type=ChannelType.GUILD_TEXT, parent_name=None):
```
asdfryan, line 43: Can we document what kinds of names are supported? (Emojis? special chars?)

What happens when a channel with the same name exists?

akirabaruah: Yep, can do. Looks like Discord supports emojis and special characters, but converts spaces to hyphens, so we can probably remove the slugify logic and directly request the name of the puzzle. WDYT?

akirabaruah: For existing channels with same names, we currently just create a new channel with the same name. To ensure channel name uniqueness, we can probably use the `get_or_create_channel()` method instead of `create_channel()` and maybe add a `prune_redundant_channels()` somewhere.

akirabaruah:
I like the idea of abstracting comms channel management with swappable backend implementations. Seems simple enough to implement as @asdfryan outlined :+1: 

We should probably keep that abstraction as simple as possible, essentially just requiring CRUD methods for some CommsChannel class/model using a string-based ID. Features like nesting within folders/categories or deduplication should probably be specific to the provider implementation. Does that sound reasonable?

rgossiaux:
> I like the idea of abstracting comms channel management with swappable backend implementations. Seems simple enough to implement as @asdfryan outlined 👍
> 
> We should probably keep that abstraction as simple as possible, essentially just requiring CRUD methods for some CommsChannel class/model using a string-based ID. Features like nesting within folders/categories or deduplication should probably be specific to the provider implementation. Does that sound reasonable?

Yeah, my instinct personally is that the interface it should expose is just some functions to handle puzzles being created/updated/deleted and then it can take care of the internals from there. Don't want to overengineer this but at least then the logic is all in one place and easier to swap out.

chat/models.py
```diff
@@ -0,0 +1,53 @@
+from django.conf import settings
+from django.db import models
+from django.dispatch import receiver
+
+
+def _get_service_choices():
+    return [(p, p) for p in settings.CHAT_SERVICES.keys()]
+
+
+class ChatRoom(models.Model):
```
asdfryan, line 10: Can you add some comments specifying what this is? i.e. it can include both text and voice etc.

chat/models.py
```diff
@@ -0,0 +1,53 @@
...
+    def get_service(self):
+        return settings.CHAT_SERVICES[self.service].get_instance()
+
+    def create_channels(self):
+        service = self.get_service()
+        self.text_channel_id = service.create_text_channel(self.name)
+        self.voice_channel_id = service.create_voice_channel(self.name)
+        self.save(update_fields=["text_channel_id", "voice_channel_id"])
+
+    def delete_channels(self):
```
asdfryan, line 36: I'm assuming that other views / models should only be accessing ChatRoom and not ChatService, right? If so, we should add some comments specifying this.

If so, I think we may need to expose more here (deleting voice channel ONLY, e.g.) but I can do this later when I implement that feature.

smallboard/settings.py
```diff
@@ -168,6 +168,16 @@
...
 
+# Discord API. See
+# https://support.discord.com/hc/en-us/articles/206346498-Where-can-I-find-my-User-Server-Message-ID-
+DISCORD_API_TOKEN = os.environ.get("DISCORD_API_TOKEN", None)
+
+# Discord server ID.
+DISCORD_GUILD_ID = os.environ.get("DISCORD_GUILD_ID", None)
+
+# Category (folder) to contain generated channels.
+DISCORD_GUILD_CATEGORY = os.environ.get("DISCORD_GUILD_CATEGORY", "puzzles")
```
asdfryan, line 179: I wonder if we should also shard this (in addition to archive). Feel free to do this in a feature PR.

discord_lib/discord_chat_service.py
```diff
@@ -0,0 +1,67 @@
...
+    def create_channel(self, name, chan_type=ChannelType.GUILD_TEXT, parent_name=None):
+        parent_id = None
+        if parent_name:
+            # Use a Discord category as the parent folder for this channel.
+            parent = self.get_or_create_channel(parent_name, ChannelType.GUILD_CATEGORY)
+            parent_id = parent.id
+        channel = self._client.guilds_channels_create(
+            self._guild_id,
+            chan_type,
+            slugify(name),
```
asdfryan, line 48: Do we need to call slugify here?

asdfryan (commented):
Generally looks great! Thanks for doing this! Added a few comments here and there.

smallboard/settings.py
akirabaruah: Yeah +1 we should probably implement some sharding scheme around this. Created #267 to track.

chat/models.py
akirabaruah: Sounds good. I'll update comments here and in the ChatService class to clarify that Django models/views should only use ChatRoom methods.

discord_lib/discord_chat_service.py
akirabaruah: No, if fact, we should probably *not* call it. Nice catch :)

chat/models.py
akirabaruah: Done!

asdfryan (approved):

PR #264, opened 2020-12-18
Conditionally disable answer queue and PENDING status

asdfryan:


rgossiaux (approved):
Nice, lgtm, thanks :) Was going to say that I was a little surprised that this didn't break tests but it looks like it did lol

asdfryan:
> Nice, lgtm, thanks :) Was going to say that I was a little surprised that this didn't break tests but it looks like it did lol

Fixed the test.

PR #269, opened 2020-12-23
[react] Puzzle tag API and UI

rgossiaux:
Add in the functionality to add/remove existing tags to a puzzle and to create new tags.

It looks a little shitty and the UX is a little shitty too but for the moment I'm just worried about the functionality.

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -29,6 +29,13 @@ answer_edit = views.AnswerViewSet.as_view(
     }
 )
 
+tag_detail = views.PuzzleTagViewSet.as_view(
+    {
+        "get": "retrieve",
+        "delete": "destroy",
+    }
+)
+
 urlpatterns = [
     path("v1/hunt/<int:pk>", views.HuntAPIView.as_view(), name="hunt_api_view"),
     path("v1/hunt/<int:hunt_id>/puzzles", puzzle_list, name="puzzle_list"),
@@ -38,5 +45,10 @@ urlpatterns = [
         "v1/hunt/<int:hunt_id>/puzzles/<int:puzzle_id>/answer/<int:pk>",
         answer_edit,
         name="answer_edit",
+    ),
+    path(
+        "v1/hunt/<int:hunt_id>/puzzles/<int:puzzle_id>/tags/<int:pk>",
+        tag_detail,
+        name="tag_detail",
     ),
 ]
--- a/api/urls.py
+++ b/api/urls.py
@@ -29,6 +29,12 @@ answer_edit = views.AnswerViewSet.as_view(
     }
 )
 
+tag_list = views.PuzzleTagViewSet.as_view(
+    {
+        "post": "create",
+    }
+)
+
 tag_detail = views.PuzzleTagViewSet.as_view(
     {
         "get": "retrieve",
@@ -45,6 +51,9 @@ urlpatterns = [
         "v1/hunt/<int:hunt_id>/puzzles/<int:puzzle_id>/answer/<int:pk>",
         answer_edit,
         name="answer_edit",
+    ),
+    path(
+        "v1/hunt/<int:hunt_id>/puzzles/<int:puzzle_id>/tags", tag_list, name="tag_list"
     ),
     path(
         "v1/hunt/<int:hunt_id>/puzzles/<int:puzzle_id>/tags/<int:pk>",
--- a/api/views.py
+++ b/api/views.py
@@ -9,7 +9,7 @@ from rest_framework import viewsets
 from answers.models import Answer
 from chat.models import ChatRoom
 from hunts.models import Hunt
-from puzzles.models import Puzzle, PuzzleModelError, PuzzleTag
+from puzzles.models import Puzzle, PuzzleModelError, PuzzleTag, is_ancestor
 from .serializers import (
     AnswerSerializer,
     HuntSerializer,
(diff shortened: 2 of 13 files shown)


akirabaruah (changes requested):
This is great! If it's not too tedious, we should probably add some basic unit tests for the API. There are some good error cases that we probably don't want to regress on.

api/urls.py
```diff
@@ -16,8 +16,29 @@
...
+tag_list = views.PuzzleTagViewSet.as_view(
+    {
+        "post": "create",
+    }
+)
+
+tag_detail = views.PuzzleTagViewSet.as_view(
+    {
+        "get": "retrieve",
+        "delete": "destroy",
```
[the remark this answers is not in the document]
rgossiaux: These names come from DRF: https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions

api/views.py
```diff
@@ -108,3 +108,70 @@ def create(self, request, **kwargs):
...
+        with transaction.atomic():
+            hunt = get_object_or_404(Hunt, pk=self.kwargs["hunt_id"])
+            puzzle = get_object_or_404(Puzzle, pk=self.kwargs["puzzle_id"])
+            serializer = self.get_serializer(data=request.data, context={"hunt": hunt})
+            serializer.is_valid(raise_exception=True)
+            (tag_name, tag_color) = (
+                serializer.validated_data["name"],
+                serializer.validated_data["color"],
+            )
+            (tag, _) = PuzzleTag.objects.update_or_create(
```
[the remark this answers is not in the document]
rgossiaux: Sure (this was actually copied from puzzles/views.py)

rgossiaux:
Yeah, I was planning on adding tests for all this/porting our existing tests over after finishing the feature work / before "launching", though there's an argument that it's better to do it while things are fresher. It's one of the checklist items in #251

akirabaruah:
Okay yeah that sounds fine, as long we're planning to add tests relatively
soon. +1 that adding tests sooner is nice, but I can understand that we
want to get the minimum features up ASAP. Thanks for clarifying the plan.

api/views.py
```diff
@@ -108,3 +108,70 @@ def create(self, request, **kwargs):
...
+                    return Response(
+                        {
+                            "detail": '"Unable to assign metapuzzle since doing so would introduce a meta-cycle."'
+                        },
+                        status=400,
+                    )
+                # the post m2m hook will add tag
+                puzzle.metas.add(metapuzzle)
+            else:
+                puzzle.tags.add(tag)
```
asdfryan, line 175: Do we need to call puzzle.save()?

hunts/src/TagCell.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import Badge from "react-bootstrap/Badge";
+import { useDispatch, useSelector } from "react-redux";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faPlus } from "@fortawesome/free-solid-svg-icons";
+import { deletePuzzleTag, fetchPuzzles } from "./puzzlesSlice";
```
asdfryan, line 6: These imports seem no longer needed?

api/views.py
rgossiaux: No, `.add` saves the relationship in the DB: https://docs.djangoproject.com/en/3.1/topics/db/examples/many_to_many/

api/views.py
```diff
@@ -6,8 +6,8 @@
 from rest_framework import viewsets
 
 from hunts.models import Hunt
-from puzzles.models import Puzzle, PuzzleModelError
-from .serializers import HuntSerializer, PuzzleSerializer
+from puzzles.models import Puzzle, PuzzleModelError, PuzzleTag, is_ancestor
```
akirabaruah, line 9: Probably beyond the scope of this PR, but shouldn't we make `is_ancestor` a method of Puzzle? Seems like `puzzle_object.is_ancestor_of(other_puzzle_object)` would make more sense.

api/views.py
```diff
@@ -108,3 +108,70 @@ def create(self, request, **kwargs):
...
+                return Response(
+                    {
+                        "detail": "You cannot remove a meta's tag (%s) from itself"
+                        % tag.name
+                    },
+                    status=400,
+                )
+            if tag.is_meta:
+                meta = Puzzle.objects.get(name=tag.name, hunt=puzzle.hunt)
+                # the post m2m hook will remove tag
```
akirabaruah, line 136: Have we considered creating m2m hooks on the PuzzleTag object that checks `is_meta` and makes these updates accordingly? That way, we could probably just call `puzzle.tags.remove(tag)` here and `puzzle.tag.add(tag)` below and avoid having these conditionals. It seems a bit strange to use the `metas` field in some cases and `tags` field in other cases, when we're really creating/deleting tag objects under the hood.

akirabaruah (approved):
LGTM overall, modulo one question about possibly simplifying the Tag creation/deletion logic 👍

api/views.py
rgossiaux: I agree, good suggestion.

api/views.py
rgossiaux: Yeah, that makes sense to me (this was just the existing structure copied over to this file), especially if we're always going to be going in the tag => meta direction.

PR #275, opened 2020-12-28
[react] Support answer deletion, editing, and adding (w/ multiple answer support)

asdfryan:
Screen Shot 2020-12-27 at 5 48 21 PM (2)
Screen Shot 2020-12-27 at 5 48 23 PM (2)
Screen Shot 2020-12-27 at 5 48 26 PM (2)
Screen Shot 2020-12-27 at 5 48 35 PM (2)
Screen Shot 2020-12-27 at 5 48 47 PM (2)
Screen Shot 2020-12-27 at 5 48 49 PM (2)


hunts/src/AnswerCell.js
```diff
@@ -1,31 +1,103 @@
...
+          )
+        }
+      >
+        Submit Answer
+      </Button>
+    );
+  }
+  return (
+    <>
+      {row.original.guesses.map(({ id, text }) => (
```
rgossiaux, line 37: The children from this `.map` need a `key` prop-- there's a warning message in the dev console about it.

hunts/src/AnswerCell.js
```diff
@@ -1,31 +1,103 @@
...
+                    huntId,
+                    puzzleId: row.values.id,
+                    answerId: id,
+                    text,
+                  },
+                })
+              )
+            }
+          >
+            <Badge pill variant="light">
```
rgossiaux, line 56: lWhat do you think about just removing these `Badge`s around the edit/delete buttons, or making them less prominent (maybe `outline-light` instead of `light`)?

hunts/src/AnswerCell.js
```diff
@@ -1,31 +1,103 @@
...
+              props: {
+                huntId,
+                puzzleId: row.values.id,
+                puzzleName: row.values.name,
+              },
+            })
+          )
+        }
+      >
+        <Badge pill variant="light">
```
rgossiaux, line 97: It kinda bugs me a bit having this button always visible, it adds some clutter. Here's what last year's mystery hunt looks like:

image

All the buttons feel like they draw my eyes away and make the answers harder to read. OTOH we definitely need edit/delete answer buttons in some form.

Not sure what the best thing to do is (I'm certainly no designer lol). What do you think of having a special "MULTIPLE ANSWERS" tag and only assuming that the puzzle takes multiple answers if it has this tag? I think nearly every puzzle we do will only have one answer so I feel like we should optimize for that case.

answers/models.py
```diff
@@ -47,6 +48,10 @@ def set_notes(self, notes):
     def get_notes(self):
         return self.response
 
+    def update(self, new_answer):
+        self.text = new_answer
+        self.save()
+
     class Meta:
```
rgossiaux, line 55: Are Answers ever ordered anywhere right now? We might want to add an ordering to the meta class to prevent any UI jank (can probably just order by `created_on`).

rgossiaux (changes requested):
Thanks! Some comments inline.

How is this going to work with the Google sheets integration? Actually now that I look at it I don't think you're handling Google sheets at all right now in your add answer flow, are you?

asdfryan:
> Thanks! Some comments inline.
> 
> How is this going to work with the Google sheets integration? Actually now that I look at it I don't think you're handling Google sheets at all right now in your add answer flow, are you?

Not handling sheets at all -- my impression was that the sheets part (changing the title to [SOLVED] was low-pri) from the last meeting. I think there's a separate issue for that.

rgossiaux:
> > Thanks! Some comments inline.
> > How is this going to work with the Google sheets integration? Actually now that I look at it I don't think you're handling Google sheets at all right now in your add answer flow, are you?
> 
> Not handling sheets at all -- my impression was that the sheets part (changing the title to [SOLVED] was low-pri) from the last meeting. I think there's a separate issue for that.

We have an integration with Google sheets to automatically sync answers from feeders to meta sheets-- see `update_meta_sheet_feeders` and callers. That's the part that's regressed.

hunts/src/AnswerCell.js
asdfryan: I agree.

IMO the best thing to do is to actually get rid of all the buttons (including edit/delete), and have the answer be a link to a modal that allows you to modify, delete, or add.

Same can be said for the name column IMO. If you agree with this, I'd prefer to do both in the same, separate PR.

api/serializers.py
```diff
@@ -40,6 +42,11 @@ class PuzzleSerializer(serializers.ModelSerializer):
     def get_tags(self, obj):
         return [{"name": tag.name, "color": tag.color} for tag in obj.tags.all()]
 
+    def get_guesses(self, obj):
```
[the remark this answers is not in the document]
asdfryan: Good point -- Done.

api/views.py
```diff
@@ -72,6 +73,47 @@ def create(self, request, **kwargs):
...
+            with transaction.atomic():
+                answer = self.get_object()
+                puzzle = get_object_or_404(Puzzle, pk=self.kwargs["puzzle_id"])
+                serializer = self.get_serializer(
+                    answer, data=request.data, partial=True
+                )
+                serializer.is_valid(raise_exception=True)
+                text = self.__sanitize_answer(serializer.validated_data["text"])
+                answer.update(text)
+                puzzle.save()
```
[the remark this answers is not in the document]
asdfryan: Thanks for the suggestion! Done.

api/views.py
```diff
@@ -72,6 +73,47 @@ def create(self, request, **kwargs):
...
+        except IntegrityError as e:
```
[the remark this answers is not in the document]
asdfryan: Thanks! Done.

api/views.py
```diff
@@ -72,6 +73,47 @@ def create(self, request, **kwargs):
 
         return Response(PuzzleSerializer(puzzle).data)
 
+    def destroy(self, request, pk=None, **kwargs):
+        puzzle = None
+        with transaction.atomic():
+            hunt = get_object_or_404(Hunt, pk=self.kwargs["hunt_id"])
```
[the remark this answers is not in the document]
asdfryan: removed.

hunts/src/AnswerCell.js
asdfryan: I tried both -- didn't seem to make a huge difference.

Regardless, I think we should just remove these buttons (in a future PR) from the answer and name columns and have all of these options in a modal.

answers/models.py
asdfryan: Done.

hunts/src/AnswerCell.js
asdfryan: I couldn't figure out how to make this disappear -- I added keys to every fragment and span inside this map.

asdfryan:
> > > Thanks! Some comments inline.
> > > How is this going to work with the Google sheets integration? Actually now that I look at it I don't think you're handling Google sheets at all right now in your add answer flow, are you?
> > 
> > 
> > Not handling sheets at all -- my impression was that the sheets part (changing the title to [SOLVED] was low-pri) from the last meeting. I think there's a separate issue for that.
> 
> We have an integration with Google sheets to automatically sync answers from feeders to meta sheets-- see `update_meta_sheet_feeders` and callers. That's the part that's regressed.

Got it. This should work now. Here's what it looks like with multiple answers: 

Screen Shot 2020-12-29 at 4 59 36 PM

api/views.py
```diff
@@ -68,10 +58,40 @@ def create(self, request, **kwargs):
                 answer.status = Answer.CORRECT
                 puzzle.answer = answer.text
                 answer.save()
+                for meta in puzzle.metas.all():
```
rgossiaux, line 61: can we wrap this in a `transaction.on_commit()` block? Been trying to avoid long-running db transactions

api/views.py
```diff
@@ -68,10 +58,40 @@ def create(self, request, **kwargs):
...
+            answer.delete()
+            # If a SOLVED puzzle has no more correct answers, revert status to SOLVING.
+            if (
+                not puzzle.guesses.filter(status=Answer.CORRECT)
+                and puzzle.status == Puzzle.SOLVED
+            ) or (not puzzle.guesses.all() and puzzle.status == Puzzle.PENDING):
+                puzzle.status = Puzzle.SOLVING
+                puzzle.save()
+
+            for meta in puzzle.metas.all():
```
rgossiaux, line 81: Here too

api/views.py
```diff
@@ -68,10 +58,40 @@ def create(self, request, **kwargs):
...
+                GoogleApiClient.update_meta_sheet_feeders(meta)
+
+        return Response(PuzzleSerializer(puzzle).data)
+
+    def partial_update(self, request, pk=None, **kwargs):
+        super().partial_update(request, pk, **kwargs)
+
+        puzzle = get_object_or_404(Puzzle, pk=self.kwargs["puzzle_id"])
+        for meta in puzzle.metas.all():
```
rgossiaux, line 90: And here

hunts/src/AnswerCell.js
rgossiaux: Ok, let's merge this and then iterate on the UI afterwards

rgossiaux (approved):
Thanks! Left minor comments but overall lgtm

api/views.py
```diff
@@ -58,8 +63,9 @@ def create(self, request, **kwargs):
                 answer.status = Answer.CORRECT
                 puzzle.answer = answer.text
                 answer.save()
-                for meta in puzzle.metas.all():
-                    GoogleApiClient.update_meta_sheet_feeders(meta)
+                transaction.on_commit(
```
rgossiaux, line 66: This doesn't work, you're missing a lambda.

PR #277, opened 2020-12-30
Handle the no-discord-API-token case

dan-simon:
What it says. I haven't added tests or checked very hard that this doesn't break other stuff and honestly what I did seemed a bit too simple.

Just to be clear, this PR is meant to, if no discord API token is in config, just not create discord channels (but otherwise do the same things).

### DIFF
--- a/hunts/views.py
+++ b/hunts/views.py
@@ -173,7 +173,7 @@ class HuntView(LoginRequiredMixin, View):
             if google_api_client:
                 google_api_client.add_puzzle_link_to_sheet(puzzle_url, sheet)
 
-            if settings.CHAT_DEFAULT_SERVICE is not None:
+            if settings.CHAT_DEFAULT_SERVICE:
                 chat_room = ChatRoom.objects.create(
                     service=settings.CHAT_DEFAULT_SERVICE, name=name
                 )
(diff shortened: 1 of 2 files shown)


asdfryan (approved):

rgossiaux (changes requested):
What's the prod_db.sql file doing there?

dan-simon:
I think it wasn't caught by .gitignore. I'll remove it.

hunts/views.py
```diff
@@ -173,10 +173,14 @@ def post(self, request, hunt_slug):
             if google_api_client:
                 google_api_client.add_puzzle_link_to_sheet(puzzle_url, sheet)
 
-            chat_room = ChatRoom.objects.create(
-                service=settings.CHAT_DEFAULT_SERVICE, name=name
-            )
-            chat_room.create_channels()
+            if settings.CHAT_DEFAULT_SERVICE is not None:
```
akirabaruah, line 176: nit: Can we simplify this to just `if settings.CHAT_DEFAULT_SERVICE:`? It's a bit nicer to read and also catches the empty string case.

akirabaruah (approved):
LGTM with a nit - thanks Dan!

rgossiaux (approved):
Thanks!

PR #278, opened 2020-12-30
[react] Reactify discord channel creation and create cell in react view

asdfryan:
Screen Shot 2020-12-30 at 3 23 54 PM (2)


api/serializers.py
```diff
@@ -29,14 +30,31 @@ class Meta:
         fields = ("text",)
 
 
+class ChatRoomSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ChatRoom
+        fields = (
+            "text_invite_url",
```
akirabaruah, line 37: Should we name these like `text_channel_url` instead to match the model fields names? I'm also thinking that non-Discord backends might not call them invite URLs, so X_channel_url seems sufficiently generalized?

chat/models.py
```diff
@@ -51,6 +51,9 @@ class ChatRoom(models.Model):
     text_channel_id = models.CharField(max_length=255, null=True, blank=True)
     audio_channel_id = models.CharField(max_length=255, null=True, blank=True)
 
+    text_invite_url = models.URLField(blank=True)
+    audio_invite_url = models.URLField(blank=True)
```
akirabaruah, line 55: Same comment as earlier: should we name these as X_channel_url to match the fields above? Also might be good to group them by text/audio.

chat/tests.py
```diff
@@ -24,6 +24,9 @@ def delete_text_channel(self, channel_id):
     def delete_audio_channel(self, channel_id):
         self.audio_channels.remove(channel_id)
 
+    def create_invite_link(self, channel_id):
```
akirabaruah, line 27: We should probably say "url" instead of "link" here, to be extra explicit about the return value.

discord_lib/discord_chat_service.py
```diff
@@ -69,3 +69,7 @@ def get_or_create_channel(self, name, chan_type=ChannelType.GUILD_TEXT):
         else:
             channel = self.create_channel(name, chan_type)
         return channel
+
+    def create_invite_link(self, channel_id):
+        invite = self._client.channels_invites_create(channel_id, max_age=0)
+        return "https://discord.gg/{}".format(invite.code)
```
akirabaruah, line 75: You might want to use a Python f-string here, like: `f"https://discord.gg/{invite.code}"` since it's fewer characters and possibly easier to read.

chat/tests.py
akirabaruah: In general, would `create_channel_url` make sense here?

hunts/src/HuntViewMain.js
```diff
@@ -66,6 +66,22 @@ const TABLE_COLUMNS = [
...
+          <a href={row.original.chat_room.audio_invite_url} target="_blank">
+            Voice
+          </a>
+          <br />
+          <a href={row.original.chat_room.text_invite_url} target="_blank">
+            Text
+          </a>
+        </>
+      ) : null,
+  },
```
akirabaruah, line 84: I love how simple this becomes in React 😄

akirabaruah (changes requested):
LGTM overall, with a question about possibly changing "invite_url" names to "channel_url"

akirabaruah:
Also, does this PR subsume #277 ?

dan-simon:
I don't think it does, since #277 also changes `settings.py`.

api/serializers.py
asdfryan: Done.

akirabaruah (approved):
Thanks!

Issue #274, opened 2020-12-27
Google Oauth Keys should not be visible in settings.py


asdfryan:

rawxfish:
No longer visible. Changed keys, see discord for them.

PR #279, opened 2021-01-01
remove google auth keys (#274)

rawxfish:


rgossiaux (approved):
Lol thank you. Did you update heroku too?

Are any of these sensitive? It looks like IDs/public keys/other OK information. Wondering if we need to rotate anything out or not.

rawxfish:
I generated new keys just to be safe and will send those around after this change

PR #281, opened 2021-01-02
Make react the default view in preparation for test tomorrow

asdfryan:


akirabaruah (approved):
LGTM 👍 thanks for flipping the switch. Excited to see how it goes

akirabaruah:
Oh are we testing tomorrow? Is a subteam using SB for the UMD hunt?

rgossiaux (approved):
I think this is fine for now in order to test tomorrow, though we still need to update the tests.

PR #282, opened 2021-01-02
Fix windows line ending issue with Docker

rgossiaux:
Git on windows automatically converts line endings to CRLF on checkout,
which makes these patch files not readable inside Docker. This should
(hopefully) exclude the patch files from this and fix the problem on
Windows though since I don't have a Windows machine I haven't actually
tested it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+patches/*.patch text eol=lf


rgossiaux:
cc @bcai2 let me know if this fixes the problem for you.

bcai2:
Windows issues when cloning with CLRF (at least those of which I've found so far):
- Attempting to install patches (the `RUN yarn install` in the Dockerfile) breaks when building. One fix was to convert the patch files' line endings and replace the `install` line with the following, applying patches individually:
`RUN patch -p1 -i patches/patch-package+6.2.2.patch`
`RUN patch -p1 -i patches/react-table+7.6.2.patch`

- Building the web applet sometimes fails with exit code 1. Googling suggests that this is another line ending issue, but I can't replicate the error (or find the page I was looking at...).

All of these seem to be circumvented by preventing automatic line conversion, which can be done on Windows by setting `core.autoclrf` to `input` (clones as LF, converts CRLF to LF upon committing) or `false` (no modifications when cloning or committing, which). This can be done with `git config --global core.autocrlf true`.

Alternatively, adding a few lines to a `.gitattributes` file should also do the trick for any future Windows users (which just prevents line conversion while cloning).

rgossiaux:
Yeah this PR adds `.gitattributes` and should (I hope) fix the patch file issues. I'm wondering what other files need to be added to it.

bcai2:
I'm not sure anything outside of that needs to be added. If any issues come up for me in the future, they're most likely not Windows-exclusive. I'll update if I do find anything else, though.

bcai2 (approved):
LGTM. I might consider changing this to `* text eol=lf` since it seemed like I was running into more line ending errors beyond patching. 
I'm not sure if that affects building manually on Windows, but I haven't tried doing so yet (and that seems like a major hassle...).

PR #288, opened 2021-01-02
Fix transaction.on_commit bugs

rgossiaux:
From c0472015a0fb54ff


asdfryan (approved):
Oops -- thanks for the fix.

PR #291, opened 2021-01-02
React: Include default tags in add tag modal

### LINKED ISSUE #286: Default tags don't appear in React UI
high priority / low priority / backsolved-- not handled right now since we're just showing all tags current assigned to any puzzle.

rgossiaux:
Fixes #286

This is a little hacky, I think. Ideally the API wouldn't let you create
tags with reserved colors. In that case the default tags need to exist
in python (passed down as part of the hunt info, probably)
and then the API validation needs to be updated.

Other things are higher pri than making this work elegantly though so
this is good enough for now.


asdfryan (approved):
Seems fine to me.

PR #293, opened 2021-01-02
Validate URLs as URLs in API

rgossiaux:
Now it will complain if you don't enter a full URL with protocol,
matching the old UI behavior. Without this change it doesn't care and
just normalizes whatever you input.

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -96,7 +96,7 @@ class PuzzleSerializer(serializers.ModelSerializer):
     tags = serializers.SerializerMethodField(required=False)
     guesses = serializers.SerializerMethodField()
     # Have to specify this explicitly for validate_url to run
-    url = serializers.CharField()
+    url = serializers.URLField()
     hunt_id = serializers.PrimaryKeyRelatedField(
         read_only=True, default=CurrentHuntDefault()
     )


rgossiaux:
Actually i think this is too strict now (it rejects `foo.com` when the old UI accepts it for example, and I think it's a reasonable input) so I'm going to iterate on it more.

erwa (approved):
Looks good. It seems you marked this PR as a "Draft"? I think you have to remove that to merge this.

asdfryan:
I see this is still in draft -- is this ready for review?

api/serializers.py
```diff
@@ -96,7 +96,7 @@ class PuzzleSerializer(serializers.ModelSerializer):
     tags = serializers.SerializerMethodField(required=False)
     guesses = serializers.SerializerMethodField()
     # Have to specify this explicitly for validate_url to run
-    url = serializers.CharField()
+    url = serializers.URLField()
```
erwa, line 99: Does this allow an empty url? I seem to recall we discussed at some point we might want to create a puzzle placeholder where the url was left blank (and perhaps filled in later).

erwa:
> Actually i think this is too strict now (it rejects `foo.com` when the old UI accepts it for example, and I think it's a reasonable input) so I'm going to iterate on it more.

I'm actually fine with require a scheme prefix, as in practice, we'll be copying and pasting URLs from the browser, and the http/https will automatically be included when copying. But if you want to support more user-friendly inputs like foo.com (which makes testing slightly easier), you could do something like auto-prepend http:// when it's missing. If it should be https://, one can always edit the URL after the puzzle is created.

Issue #250, opened 2020-12-06
Make puzzle creation fast


asdfryan:
Bottleneck is currently in sheet creation. We should try creating the sheet asynchronously (and have a manual workaround to set the sheet if it fails).

rgossiaux:
We might want to set up something like Celery. Then we can use it for various async tasks: Google API calls, Discord API calls, periodically running jobs (like cleaning up Discord channels).

I also really like @rawxfish 's idea of keeping around a few empty sheets that we can rename, instead of having to wait to copy a sheet before assigning it to a puzzle.

PR #295, opened 2021-01-02
setup celery to make puzzle creation fast (#250)

rawxfish:
works locally, need to test on heroku and with more edge cases.


google_api_lib/drive.py
```diff
@@ -0,0 +1,245 @@
...
+from googleapiclient import _auth
+
+from celery import shared_task
+from .utils import GoogleApiClientTask
+from puzzles.models import Puzzle
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task(base=GoogleApiClientTask, bind=True)
```
rgossiaux, line 16: I think the convention is to collect all delayed tasks in a `/tasks.py` file (one per django app) so they're easier to find

smallboard/settings.py
```diff
@@ -64,6 +64,7 @@
     "social_django",
     "taggit",
     "rest_framework",
+    "django_celery_results",
```
rgossiaux, line 67: What does this do / are we using the task results anywhere? I think it's also possible to just throw away task results if we don't need them

puzzles/models.py
```diff
@@ -138,16 +135,6 @@ def update_metadata(self, new_name, new_url, new_is_meta):
...
-        def update_sheets():
-            if is_new_url:
-                google_api_client = GoogleApiClient.getInstance()
-                if google_api_client:
-                    google_api_client.add_puzzle_link_to_sheet(self.url, self.sheet)
-            for meta in self.metas.all():
-                GoogleApiClient.update_meta_sheet_feeders(meta)
-
-        transaction.on_commit(update_sheets)
-
```
rgossiaux, line 150: Why deleting this?

smallboard/settings.py
rgossiaux: Oh ok, I see that `get_file_user_emails` returns a result, seems reasonable.

rgossiaux (commented):
Yay!! Very excited.

I think we'll also want to update the docker configs. I can help with that though I think it's pretty straightforward, just need to add a few more containers to docker-compose.yaml

google_api_lib/drive.py
```diff
@@ -0,0 +1,245 @@
+import httplib2
```
erwa, line 1: Is it possible to do `git mv google_api_client.py drive.py` as a separate commit so it's easier to see what the changes are to the file and to preserve the lineage? (This can be part of this PR, just as a separate commit before making the changes to the file. I think if done this way, GitHub should show "google_api_client.py -> drive.py" and show the diff.)

google_api_lib/drive.py
rawxfish: Hmm yeah I actually wasn't sure what the best thing to do was b/c I had to split the tasks between drive.py, which depends on Puzzle model to update google sheets url, and whitelist.py which is called in settings.py before Puzzle model exists yet to get the whitelisted emails. The whitelisted emails should really be removed from there and be made a field of a hunt. I was gonna wait to do that later, but maybe I should just do it as apart of this....

puzzles/models.py
rawxfish: google_api_lib now directly updates Puzzles, so I moved all calls to google_api_lib to api/views.py to avoid circular dependencies. This part was moved to https://github.com/cardinalitypuzzles/smallboard/blob/2fe8a5d0e11f39156e56a51d30c8a49f7770e5d4/api/views.py#L160-167

google_api_lib/drive.py
rgossiaux: Aha yeah. If the one in whitelist.py is needed at startup then is there much point making it asynchronous to begin with? Maybe we can just leave it synchronous for now and move the other tasks to task.py? We definitely need to improve the permissions modeling at some point but it feels it can wait for now

rawxfish: It's still being called synchronously there. That's a good point though, I'll rename drive.py into task.py and either keep whitelist.py or rename that to sync_task.py or something lol.

api/views.py
```diff
@@ -151,14 +151,24 @@ def partial_update(self, request, pk=None, **kwargs):
...
                 if "status" in data:
                     puzzle.status = data["status"]
                     puzzle.save()
+
+                if is_new_url and google_api_lib.enabled():
+                    google_api_lib.task.add_puzzle_link_to_sheet.delay(
+                        new_url, puzzle.sheet
+                    )
+                if puzzle.is_meta:
+                    google_api_lib.task.update_meta_sheet_feeders.delay(puzzle.id)
```
rgossiaux, line 170: Actually could we wrap these in `transaction.on_commit` ? If the celery tasks run before the transaction commits (or if it doesn't commit successfully) the task is going to error.

api/views.py
```diff
@@ -192,14 +197,14 @@ def create(self, request, **kwargs):
...
 
-            if google_api_client:
-                transaction.on_commit(
-                    lambda: google_api_client.add_puzzle_link_to_sheet(
-                        puzzle_url, sheet
-                    )
+            if google_api_lib.enabled():
+                google_api_lib.task.create_google_sheets.delay(
+                    puzzle.id, name, puzzle_url
                 )
```
rgossiaux, line 205: Ditto here

api/views.py
```diff
@@ -192,14 +197,14 @@ def create(self, request, **kwargs):
...
+            puzzle = serializer.save(hunt=hunt, chat_room=chat_room)
 
-            if google_api_client:
-                transaction.on_commit(
-                    lambda: google_api_client.add_puzzle_link_to_sheet(
-                        puzzle_url, sheet
-                    )
+            if google_api_lib.enabled():
+                google_api_lib.task.create_google_sheets.delay(
+                    puzzle.id, name, puzzle_url
```
rgossiaux, line 204: Also here

rgossiaux (approved):
Lgtm, thanks!!

PR #297, opened 2021-01-04
Rename Tags column to Tags/Metas in react UI per feedback from UMD test

asdfryan:
Screen Shot 2021-01-04 at 12 44 53 PM


rgossiaux (approved):
Sure, seems fine to me

Issue #219, opened 2020-11-29
Track Puzzle age / solve time


akirabaruah:
Add fields to the Puzzle model to track:

- `created_at` time
- `updated_at` time
- `solved_at` time

akirabaruah:
These field should be user-sortable in the main puzzle table.

asdfryan:
The fields have been added, but the tables are not sortable, right?

bcai2:
Not yet, as of now.

bcai2:
Come to think of it, it would be tricky to make this work in a way that preserves the meta structure that we already have. 

We could sort each "meta forest layer" by time (e.g., earliest to latest created metas, and then earliest to latest of the puzzzles within each meta), or we could just discard the meta tree structure altogether when sorting by time.

rgossiaux:
I agree that sorting by these with the tree structure might be kinda tricky. Not sure what react-table would do by default, if it would sort each level of the tree or sort everything together. 

Maybe we could display them in the puzzle table (or since space is at a premium, maybe just unlock time, or alternatively one single time column that shows unlock time for unsolved puzzles and solve time for solved ones though this could get messy), using relative timestamps (1m ago, 4h ago, etc). I think that'd be pretty useful for solvers (who want to know how old/new various puzzles are).

rgossiaux:
Let's track any UI changes at #459 . Closing this since the model code is there.

PR #298, opened 2021-01-04
Adding puzzle timestamps (#219)

bcai2:
* Added `created_on`, `updated_on`, and `solved_on` fields to Puzzles model.
* `updated_on` updates whenever the puzzle's data is modified.
* `solved_on` is set to the time when the first answer is entered. If no answer is present or the answer is cleared, this becomes `None`.


puzzles/migrations/0024_auto_20210104_2026.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.1.4 on 2021-01-04 20:26
+
+from django.db import migrations, models
+import django.utils.timezone
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('puzzles', '0023_auto_20201221_0030'),
```
asdfryan, line 10: Should run black on this -- otherwise presubmit will complain about the single quotes.

puzzles/models.py
```diff
@@ -151,6 +156,7 @@ def update_sheets():
     def set_answer(self, answer):
```
asdfryan, line 156: The answer field is now defunct in the React version because this field assumes single answer per puzzle. We now use the related model field "guesses" (see answers/) for this.

We should actually remove this function once the non-react view is deprecated.

puzzles/models.py
```diff
@@ -40,6 +41,10 @@ class Puzzle(models.Model):
     )
     url = models.URLField(blank=True)
 
+    created_on = models.DateTimeField(auto_now_add=True)
+    updated_on = models.DateTimeField(auto_now=True)
+    solved_on = models.DateTimeField(default=None, null=True, blank=True)
```
asdfryan, line 46: How is "solved_on" going to be set? In my mind, it can either be:
1) the "created_on" timestamp of the last Answer object for this puzzle that is marked as CORRECT
2) the "updated_on" timestamp of the last Answer object for this puzzle that is marked as CORRECT

On a related note, if we're just using one of the Answer timestamps, I wonder if having this as a model object is really worth it (we'd be putting the same information in 2 different places in the DB). It might just be better to have a "solved_time()" method that looks at the answers for the puzzle.

asdfryan (changes requested):
Added a few comments.

puzzles/models.py
bcai2: Option 1 is probably the way to go; I can't imagine that option 2 would be different unless we make a typo, in which case option 1 contains the time we want.  (Unless we enter in a "partial", but that wouldn't be marked as correct, and the eventual correct answer would be a different guess anyway.)

That sounds really sensible, so I'll do that instead.

puzzles/models.py
bcai2: `set_status` in `answers/models.py` references `set_answer` in `puzzles/models.py`; if we're deprecating the latter, would we have to change how the puzzle is marked as correct? 

(Is the eventual plan to have a choice for puzzle correctness that users manually change? This might be unnecessary and we can just have an informal policy to only enter correct answers, especially if this is too complicated or clunky UI-wise.)

bcai2:
All fixes should be taken care of with the latest commits (which I'm trying to squash, but hopefully the PR can take care of that?).

asdfryan (approved):

puzzles/models.py
```diff
@@ -169,6 +173,14 @@ def clear_answer(self, guess):
     def is_solved(self):
         return self.status == Puzzle.SOLVED
 
+    def solved_time(self):
+        if not self.is_solved():
+            return None
+        solved_times = [
+            answer.created_on for answer in self.guesses.filter(status=Answer.CORRECT)
+        ]
+        return max(solved_times, default=None)
```
rgossiaux, line 182: You should do this in the ORM: https://docs.djangoproject.com/en/dev/ref/models/querysets/#latest

rgossiaux (commented):
Looks like I can't request changes any more

PR #299, opened 2021-01-04
Fix bug related to discord channel deletion

asdfryan:
We store channel_id as string, but discord API returns/expects int.


akirabaruah (approved):
Nice catch. Thanks!

PR #301, opened 2021-01-05
Rewrite Puzzle solved_times using ORM

bcai2:
assuming that we maintain that is_solved returns true iff at least one correct answer is present

### DIFF
--- a/puzzles/models.py
+++ b/puzzles/models.py
@@ -176,7 +176,9 @@ class Puzzle(models.Model):
     def solved_time(self):
         if not self.is_solved():
             return None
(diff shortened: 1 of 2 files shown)


asdfryan (commented):
LGTM. Forgot to comment this on the other PR, but can you add a unit test for this?

bcai2:
I've included the unit test under `puzzles/tests.py`, which seems more sensible than including it in `answers/tests.py`, but let me know if I'm wrong about that.

rgossiaux (approved):
Thanks!

PR #300, opened 2021-01-05
Archive discord channels upon solve

asdfryan:
This does not include category sharding, which will be done in a future PR.

PR #305, opened 2021-01-06
Add initial batch of unit tests for the React API

rgossiaux:
These could be improved / made more complete but I think it's a start, and it's more testing than the old endpoints had anyway.

### DIFF
--- a/api/tests.py
+++ b/api/tests.py
@@ -1,0 +1,189 @@
+from .test_helpers import SmallboardTestCase
+from rest_framework import status
+from puzzles.models import Puzzle, PuzzleTag
+
+TEST_URL = "https://smallboard.test/"
+
+
+class ApiTests(SmallboardTestCase):
+    def test_get_hunt(self):
+        response = self.get_hunt()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(
+            response.data,
+            {
+                "id": self._hunt.pk,
+                "name": self._hunt.name,
+                "url": self._hunt.url,
+                "active": self._hunt.active,
+            },
+        )
+
+    def test_create_invalid_puzzle(self):
+        # Missing name
+        response = self.create_puzzle(
+            {"url": "https://smallboard.test", "is_meta": False}
+        )
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        # Missing url
+        response = self.create_puzzle({"name": "test name", "is_meta": False})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        # Empty string name
+        response = self.create_puzzle({"name": "", "url": TEST_URL, "is_meta": False})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        # Malformatted URL (TODO)
+        # response = self.create_puzzle(
+        #     {"name": "test name", "url": "bad_url", "is_meta": False}
+        # )
+        # self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        self.assertEqual(Puzzle.objects.count(), 0)
+
+    def test_create_puzzle(self):
+        response = self.create_puzzle({"name": "test name", "url": TEST_URL})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        puzzle = Puzzle.objects.get()
+        self.assertEqual(
+            response.data,
+            {
+                "id": puzzle.pk,
+                "name": puzzle.name,
+                "hunt_id": self._hunt.pk,
+                "url": TEST_URL,
+                "notes": "",
+                "sheet": None,
+                "chat_room": {},
+                "status": "SOLVING",
+                "tags": [],
+                "guesses": [],
+                "metas": [],
+                "feeders": [],
+                "is_meta": False,
+            },
+        )
+
+    def test_delete_puzzle(self):
+        self.create_puzzle({"name": "test name", "url": TEST_URL})
+        puzzle = Puzzle.objects.get()
+        response = self.delete_puzzle(puzzle.pk)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, {})
+        self.assertEqual(Puzzle.objects.count(), 0)
+
+    def test_edit_puzzle(self):
+        self.create_puzzle({"name": "test name", "url": TEST_URL})
+        puzzle = Puzzle.objects.get()
+
+        response = self.edit_puzzle(puzzle.pk, {"name": "new name"})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["name"], "new name")
+        puzzle.refresh_from_db()
+        self.assertEqual(puzzle.name, "new name")
+
+        response = self.edit_puzzle(puzzle.pk, {"url": "https://newurl.test/"})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["url"], "https://newurl.test/")
+        puzzle.refresh_from_db()
+        self.assertEqual(puzzle.url, "https://newurl.test/")
+
+        response = self.edit_puzzle(puzzle.pk, {"is_meta": True})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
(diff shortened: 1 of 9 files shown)


rgossiaux:
I think the test failure is legitimate -- @asdfryan in #300 it looks like you didn't handle the case when `puzzle.chat_room` is None

erwa (commented):
I haven't looked closely at the tests you added, but just a thought: do we want to add JavaScript tests that directly test the React/Redux code? I think that would add additional robustness on top of testing the Python code.

rgossiaux:
Eventually, yes. I don't want to spend the time on them before mystery hunt though because I think there are other more critical things to do, but I'm hoping to get to it after this mystery hunt. Added #307 to track.

api/tests.py
```diff
@@ -0,0 +1,193 @@
+from .test_helpers import SmallboardTestCase
+from rest_framework import status
+from puzzles.models import Puzzle, PuzzleTag
+
+if "unittest.util" in __import__("sys").modules:
+    # Show full diff in self.assertEqual.
+    __import__("sys").modules["unittest.util"]._MAX_LENGTH = 999999999
```
erwa, line 7: This seems a little strange to include here. I think you can just add `-vv` when running `./manage.py test` to get more info including the full diff. I think it'd be better to increase the verbosity when running tests so it applies to all tests rather than making the change just for these api tests.

erwa (commented):
Thanks for adding tests.  Had a few comments.

asdfryan:
> I think the test failure is legitimate -- @asdfryan in #300 it looks like you didn't handle the case when `puzzle.chat_room` is None

Oops -- will fix this.

hunts/src/api.js
```diff
@@ -21,12 +21,12 @@ function handleErrors(response) {
 }
 
 function getPuzzles(huntId) {
-  const puzzlesApiUrl = `/api/v1/hunt/${huntId}/puzzles`;
+  const puzzlesApiUrl = `/api/v1/hunts/${huntId}/puzzles`;
```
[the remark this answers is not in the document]
rgossiaux: Yes. Actually I think the URL scheme is kind of bad, there's no need for the /hunts/:id/ prefix on most of the endpoints

api/tests.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Empty string name
+        response = self.create_puzzle({"name": "", "url": TEST_URL, "is_meta": False})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        # Malformatted URL (TODO)
+        # response = self.create_puzzle(
+        #     {"name": "test name", "url": "bad_url", "is_meta": False}
+        # )
+        # self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
rgossiaux: Yeah, that was the idea / why I included this. If I merge this PR first (likely) I'll update the other one to uncomment this.

api/tests.py
rgossiaux: Oh nice, I didn't know about -v. But I think that doesn't prevent the test output from being truncated:

```
======================================================================
FAIL: test_create_puzzle (api.tests.ApiTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/smallboard/api/tests.py", line 49, in test_create_puzzle
    self.assertEqual(
AssertionError: {'id'[137 chars]G', 'tags': [], 'guesses': [], 'metas': [], 'f[25 chars]alse} != {'id'[137 chars]G', 'guesses': [], 'metas': [], 'feeders': [],[13 chars]alse}
```

Since this runs during import I believe it applies to all tests as long as you're including `api/tests.py` but it wouldn't apply if you ran eg `test answers`, so I could move it somewhere else (maybe api/test_helpers.py?)

erwa: What if you use two `v`s as in literally `-vv` for extra verbosity?

rgossiaux:
```
usage: manage.py test [-h] [--noinput] [--failfast] [--testrunner TESTRUNNER] [-t TOP_LEVEL]
                      [-p PATTERN] [--keepdb] [-r] [--debug-mode] [-d] [--parallel [N]]
                      [--tag TAGS] [--exclude-tag EXCLUDE_TAGS] [--pdb] [-b]
                      [-k TEST_NAME_PATTERNS] [--version] [-v {0,1,2,3}]
                      [--settings SETTINGS] [--pythonpath PYTHONPATH] [--traceback]
                      [--no-color] [--force-color]
                      [test_label [test_label ...]]
manage.py test: error: argument -v/--verbosity: invalid int value: 'v'
```

The previous comment was with the most permissive verbosity level (3).

api/tests.py
```diff
@@ -0,0 +1,193 @@
...
+        )
+
+    def test_delete_puzzle(self):
+        self.create_puzzle({"name": "test name", "url": TEST_URL})
+        puzzle = Puzzle.objects.get()
+        response = self.delete_puzzle(puzzle.pk)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, {})
+        self.assertEqual(Puzzle.objects.count(), 0)
```
asdfryan, line 79: Another edge case worth testing: Trying to delete a puzzle that has already been deleted

api/tests.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def test_edit_puzzle(self):
+        self.create_puzzle({"name": "test name", "url": TEST_URL})
+        puzzle = Puzzle.objects.get()
+
```
asdfryan, line 84: Edge cases: 
- Editing more than 1 field at a time
- Editing a non-editable field
- New name matches a puzzle already in DB

api/tests.py
```diff
@@ -0,0 +1,193 @@
...
+        self.assertEqual(puzzle.status, Puzzle.SOLVED)
+        self.assertEqual(len(puzzle.correct_answers()), 1)
+
+        response = self.delete_answer(puzzle.pk, guesses[1].pk)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        puzzle.refresh_from_db()
+        self.assertEqual(puzzle.status, Puzzle.SOLVING)
+        self.assertEqual(len(puzzle.correct_answers()), 0)
+
+    def test_edit_answer(self):
```
asdfryan, line 160: Edge cases:
- Editing an answer to another answer that already exists
- Special characters / emojis

hunts/src/api.js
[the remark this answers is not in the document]
rgossiaux: I'm gonna update this in another PR after this because otherwise I think the merge conflicts will get a little messy. But I think I want to move:

/hunts/:id/puzzles/:id/tags => /puzzles/:id/tags
/hunts/:id/puzzles/:id/tags/:id => /puzzles/:id/tags/:id
/hunts/:id/puzzles/:id/answers => /puzzles/:id/answers
/hunts/:id/puzzles/:id/answers/:id => /puzzles/:id/answers/:id

Hopefully that's a little more readable. Then I'll change this api.js file to use a common /api/v1/ prefix

asdfryan (approved):
Thanks! I commented some other edge cases worth testing. Feel free to merge this in first though. I can add those later.

api/tests.py
rgossiaux: We should definitely test these but I think the answers model tests do/should take care of these ones

rgossiaux: (though from the API side I think ideally we should still test that the API returns reasonable responses in these cases)

api/tests.py
erwa: I see... I am surprised at unittest's behavior.
I wonder if you can move this unittest configuration setup to the root directory and have it always be picked up?
If not, I would prefer not changing the MAX_LENGTH at all, since I feel this sort of configuration should be done globally (and put in a global place if so) rather than done in a specific test, even if it applies globally. But if you feel strongly for including it, I'm fine with keeping it, too.

hunts/src/api.js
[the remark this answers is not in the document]
erwa: Ok, that's fine with me

api/tests.py
rgossiaux: I'll just delete it, probably simplest that way.

api/tests.py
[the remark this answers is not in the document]
erwa: Oops, I linked to the wrong PR before, I meant #293

Issue #304, opened 2021-01-05
Adding starting/ending time fields when creating hunt


bcai2:
This would help out a lot for making the progress chart (#200) and displaying stats. 

It would be nice if times were able to be entered by the user when creating the hunt, which would probably require a React date and time picker.

There should also be a default behavior: if no time is entered, the start time should be assumed to be when the hunt was created. The end time is a bit ambiguous; one option is to make it some arbitrary point in the future (e.g., 10 days?).

bcai2:
Implemented by #306. The UI could be a bit better, but that can be covered in another issue.

Issue #200, opened 2020-11-26
Burn down chart visualization


akirabaruah:
@cs-miller brought up the valuable feature of visualizing a team's Hunt progress as a [burn down chart](https://en.wikipedia.org/wiki/Burn_down_chart). This visualization can help keep moral high as we make progress on a hunt.

It can also serve as visual metric of our average team solve rate as the hunt progresses.

bcai2:
Would a "build-up"-type chart be more feasible? It would be hard to set a starting point for a burn-down chart if we don't know exactly how many puzzles the hunt's going to contain or that we'll encounter.

(Otherwise, we can just make an adjustable estimate initially set to 200 or so.)

I'm thinking something along the lines of http://www.mit.edu/~puzzle/2020/statistics/, with visible meta solves as well.

bcai2:
Outline of what I'm thinking for this:

- [x] Timestamps (#219, sort of) 
- [ ] Basic stats page
  - [ ] Total puzzles completed
  - [ ] Total puzzles unsolved
  - [ ] Total puzzles open
  - [ ] Integrate this with the applet
- [ ] Progress bar line graph
  - [ ] Make graph
  - [ ] Integrate with stats page

bcai2:
The page has been implemented! (#378)
Marking this as resolved and merging the updating into the update issue (#390).

PR #306, opened 2021-01-06
Add start/end times

bcai2:
* Start/end times (#304) can now be entered when making a hunt, under `YYYY-MM-DD` or `MM/DD/YYYY` format for dates and `HH:MM` format for times. Dates/times should be entered in the default time zone (ET).
* ~~Fields can be left blank for default times. If both fields are left blank, the start time is set to the hunt creation time. If the end field is left blank, the end time is set to 3 days after the start time.~~ (No explicit default start/end times are given; see below convo. If times are not specified, substitute times are used when computing stats and making the progress chart; see #200 for more info.)
* Invalid date/time inputs are marked as invalid.
* Correctly formatted start/end times that are invalid are marked as such. This includes:
  * entering the end time while leaving the start time blank
  * entering an end time that comes before the entered start time.


rgossiaux:
Start times make sense to me, every hunt has these announced well in advance.

Do we need the end_times field? Feels like something we might not always know & also feels a little duplicative with the active status of a hunt. This is motivated by graphing progress, right? Wonder if it suffices to just look at the time of the last guess or something.

bcai2:
The end_time field could be useful for stats, like knowing when to keep computing stats when the hunt's still active and cutting off stats computation when the hunt ends, but having a fillable field probably isn't essential, especially if there's a way to mark hunts as finished. We could just mark that time as the end_time and assume that the hunt's still progressing if it hasn't been marked as finished yet. (Speaking of this, has this been implemented yet?)

I'd be okay with getting rid of it if it's deemed unnecessary.

asdfryan:
I think the graph should show [start_time, current_time] during the hunt, and [start_time, min(end_time, finish_time)] after the hunt ends, where finish_time is defined as the solve_time of the very last puzzle (or maybe last meta, but I'm not sure how to encapsulate that information -- maybe tags?).

For this reason, I agree that having the end_time is useful.

asdfryan (commented):
Can you share a screenshot of what the new hunt creation dialog box looks like?

bcai2:
image

It isn't the cleanest (an interactive picker would be really neat, but I don't think it's necessary nor will I have time to implement it), but it behaves as expected, as far as I can tell.

asdfryan:
> image
> 
> It isn't the cleanest (an interactive picker would be really neat, but I don't think it's necessary nor will I have time to implement it), but it behaves as expected, as far as I can tell.

Can we use one of the existing widgets? (e.g. https://jquense.github.io/react-widgets/api/DateTimePicker/)

bcai2:
That looks neat and simple enough to implement! It's probably best to move on for now, but I'll try to come back to it before the playtest.

hunts/models.py
```diff
@@ -22,6 +26,15 @@ def save(self, *args, **kwargs):
             self.slug = slugify(self.name)
         super().save(*args, **kwargs)
 
+    def clean(self):
+        # adds default start and end times if not provided
+        new_start_time = self.start_time
```
asdfryan, line 31: Why not just:

```
if not self.start_time:  
    self.start_time = self.created_on  
if not self.end_time:
    self.end_time = self.start_time + timedelta(days=3)
```

hunts/models.py
```diff
@@ -22,6 +26,15 @@ def save(self, *args, **kwargs):
             self.slug = slugify(self.name)
         super().save(*args, **kwargs)
 
+    def clean(self):
```
asdfryan, line 29: I think it is cleaner to override the `save()` method so you can avoid calling `save()`, `clean()`, and `save()` again in succession.

asdfryan (changes requested):
A few comments.

hunts/models.py
bcai2: The code yells at me because the fields might be empty, so start_time + timedelta might be undefined. This way, all the changes are committed at once. It's kinda hacky, but I couldn't find a cleaner way to do it.

hunts/views.py
```diff
@@ -35,7 +35,14 @@ def index(request):
...
         form = HuntForm(request.POST)
         if form.is_valid():
-            hunt = Hunt(name=form.cleaned_data["name"], url=form.cleaned_data["url"])
+            hunt = Hunt(
+                name=form.cleaned_data["name"],
+                url=form.cleaned_data["url"],
+                start_time=form.cleaned_data["start_time"],
+                end_time=form.cleaned_data["end_time"],
+            )
+            hunt.save()
```
asdfryan, line 44: See my comment above about overriding save() instead of clean(). You can also connect to the `pre_save` signal as well.

https://stackoverflow.com/questions/4380879/django-model-field-default-based-off-another-field-in-same-model

hunts/models.py
asdfryan: I'm confused -- the code snippet in my previous comment should be functionally the same as what you've written (sans a temp variable which I *thought* was unnecessary). Am I missing something?

Even if `start_time` is undefined, it should be guaranteed to be defined by the last line.

bcai2: Weirdly, `start_time` didn't seem to be defined until `save` was called (after `clean` finished). One fix I was thinking of was calling `save` again between confirming `start_time` and `end_time`, but that seems messy.

bcai2: Here's a fix. It's still a bit messy, but it's cleaner than before. `clean` doesn't have to be called, and `save` only needs to be called once.

```
    def save(self, *args, **kwargs):
        if not self.slug:
            self.slug = slugify(self.name)
        super().save(*args, **kwargs)

        # adds default start and end times if not provided
        if not self.start_time:
            self.start_time = self.created_on
            super().save(*args, **kwargs)
        if not self.end_time:
            self.end_time = self.start_time + timedelta(days=3)
            super().save(*args, **kwargs)
```

hunts/models.py
```diff
@@ -22,6 +26,14 @@ def save(self, *args, **kwargs):
             self.slug = slugify(self.name)
         super().save(*args, **kwargs)
 
+        # adds default start and end times if not provided
+        if not self.start_time:
+            self.start_time = self.created_on
+            super().save(*args, **kwargs)
+        if not self.end_time:
+            self.end_time = self.start_time + timedelta(days=3)
```
asdfryan, line 34: Actually this is probably best implemented as a `post_save()` method: https://docs.djangoproject.com/en/3.1/ref/signals/#post-save

It even has a `created` param which you can use so it only triggers on new Hunt creations.

rgossiaux:
> I think the graph should show [start_time, current_time] during the hunt, and [start_time, min(end_time, finish_time)] after the hunt ends, where finish_time is defined as the solve_time of the very last puzzle (or maybe last meta, but I'm not sure how to encapsulate that information -- maybe tags?).
> 
> For this reason, I agree that having the end_time is useful.

This isn't really a big deal but I'm still not sure I like this. For the main case of MH specifically we don't know in advance when it will end and for other hunts I think the official end times are irrelevant for us / we only care about when we finish. Maybe I agree that we should be able to track a timestamp for when we are "finished" with the hunt and the issue is just (1) I don't like specifying it at hunt creation and (2) I don't like the duplication with `hunt.active`. 

Maybe we can remove the `hunt.active` field and simply use the timestamps? And then eventually add some action for marking a hunt as "complete" to populate the end time field (not critical, for now we could use the admin console).

rgossiaux:
> Can we use one of the existing widgets? (e.g. https://jquense.github.io/react-widgets/api/DateTimePicker/)

FYI the hunt creation view is still a django template (not react), as is everything besides the main puzzles page.

asdfryan:
> > I think the graph should show [start_time, current_time] during the hunt, and [start_time, min(end_time, finish_time)] after the hunt ends, where finish_time is defined as the solve_time of the very last puzzle (or maybe last meta, but I'm not sure how to encapsulate that information -- maybe tags?).
> > For this reason, I agree that having the end_time is useful.
> 
> This isn't really a big deal but I'm still not sure I like this. For the main case of MH specifically we don't know in advance when it will end and for other hunts I think the official end times are irrelevant for us / we only care about when we finish. Maybe I agree that we should be able to track a timestamp for when we are "finished" with the hunt and the issue is just (1) I don't like specifying it at hunt creation and (2) I don't like the duplication with `hunt.active`.
> 
> Maybe we can remove the `hunt.active` field and simply use the timestamps? And then eventually add some action for marking a hunt as "complete" to populate the end time field (not critical, for now we could use the admin console).

I was actually thinking specifically about the MITMH case when I thought about this, as it does have an official end time (10AM ET Monday). Our team is unlikely to finish by that time, and I suspect we will continue to work until then (even if another team has already found the coin), so this is a reasonable cutoff time for progress tracking.

For other hunts, I do agree that the initially populated end time is more or less meaningless and will be ignored.

Nevertheless, I think keeping both fields completely optional is fine. If `start_time` is missing, we can just disable the progress page feature. I'm not entirely sure it's a good idea to default `start_time` to `created_on`.

rgossiaux:
I think in Mystery Hunt normally it's not really totally clear when they'll
close HQ, right? The hunt may run longer or shorter depending on whether
the coin has been found or how teams are progressing, or the time/energy of
the organizers. That's more what I meant.

asdfryan:
Maybe this discussion is more pedantic than useful, but in the universe of
Puzzle Hunts, I don't think it is unreasonable to imagine a team working on
a hunt with pre-defined start and end times, where the team will be
spending the entire time working, and not manage to finish.

I can see that it is debatable that 2021 MITMH may or may not fall into
that category. But my guess is that at least *some* members of Cardinality
will likely keep working on puzzles until 10AM ET Monday, even if they
close HQ events.

hunts/signals.py
```diff
@@ -0,0 +1,16 @@
+from .models import Hunt
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from datetime import timedelta
+
+
+@receiver(post_save, sender=Hunt, dispatch_uid="add_default_times")
```
asdfryan, line 7: Sorry about the continued back and forth.

After thinking more about the bigger picture / thread with Ryan G, I think it's better to not set any defaults here for either `start_time` nor `end_time`. I don't think hunt creation time is actually very good proxy for hunt start time. Nor do I think that start_time + 3 days would be a good proxy for hunt start time.

For the progress page feature, we can do it as follows:
- If `start_time` is specified, use that as the beginning of the chart. Otherwise use the earliest puzzle creation timestamp.
- If `end_time` is specified, use `max(end_time, current_time)` as the end of the chart. Otherwise use `current_time`. 

@rgossiaux for FYI

rgossiaux: I don't want to hold this up too much, I think I said my piece and beyond that I trust you guys to do what you think is best. Ultimately this is pretty minor and it should be easy to change this stuff later if it becomes a problem anyway.

Your suggestion seems reasonable to me, I especially like the earliest puzzle creation timestamp fallback which seems like a good idea.

bcai2: Sounds good to me. Choosing which time to use when start/end times are unspecified can be done when computing stats + making the chart, so I'll go ahead and implement that over there.

asdfryan (approved):
Thanks for being patient! LGTM.

asdfryan:
Looks like coverage is complaining -- is it possible to add some tests for forms.py?

bcai2:
Sure. I'll have to merge those in eventually with the HuntTests I've written for the stats page, but that shouldn't be much of a problem.

PR #312, opened 2021-01-07
Add section in dev-guide to setup .env

erwa:
Added a reminder in dev-guide.md to set up `.env`, which is required by our Docker and manual setups.

```
$ docker-compose up
ERROR: Couldn't find env file: /home/erwaman/github/cardinalitypuzzles/smallboard/.env
```

### DIFF
--- a/dev-guide.md
+++ b/dev-guide.md
@@ -34,7 +34,7 @@ Both Docker and manual setup read environment variables from an `.env` file. An 
 cp .env.test .env
 ```
 
-See below for more details on the `.env` file.
+See below for more details on the `.env` file. You probably want to set at least `DJANGO_SECRET_KEY` so that your session isn't invalidated each time you restart the web container.
 
 
 #### Docker setup
--- a/dev-guide.md
+++ b/dev-guide.md
@@ -16,8 +16,7 @@ To run Small Board manually, you will need:
 - Pipenv
 - [NodeJS](https://nodejs.org/en/download/package-manager/) (version 10 or later)
 - [Yarn](https://classic.yarnpkg.com/en/docs/install/)
-- # a local database (PostgreSQL or SQLite)
-- a local database (e.g.: Postgres)
+- a local database (e.g.: PostgreSQL or SQLite)
 
 #### Checking out the code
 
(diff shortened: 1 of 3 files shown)


dev-guide.md
```diff
@@ -27,6 +26,17 @@ To check out the code, you first need to [install Git](https://git-scm.com/book/
...
 
+### Set up an `.env` file
+
+Both Docker and manual setup read environment variables from an `.env` file. An example `.env.test` file is provided, which you should copy to `.env`:
+
+```
+cp .env.test .env
+```
+
+See below for more details on the `.env` file.
```
asdfryan, line 37: Should we put more details on how to get the necessary API keys? (Google, Discord)

Like -- "Contact these people if you want API keys" or something

rgossiaux:
Oops, thanks! For running with docker it's enough to just `touch .env` but we do assume the file exists.

We should include something about DJANGO_SECRET_KEY somewhere actually, people should set that or else they'll have annoying issues around login/session invalidation.

Maybe we should update `.env.test` to include Docker usage (in which case you don't need to set DATABASE_URL and indeed it will be ignored if you do).

dev-guide.md
erwa: Good idea, I'll add something about that. (I was planning to just grab all the keys from our Heroku deployment :)

erwa:
> Oops, thanks! For running with docker it's enough to just `touch .env` but we do assume the file exists.

That's true

> We should include something about DJANGO_SECRET_KEY somewhere actually, people should set that or else they'll have annoying issues around login/session invalidation.

Sure, I can add something about DJANGO_SECRET_KEY

> Maybe we should update `.env.test` to include Docker usage (in which case you don't need to set DATABASE_URL and indeed it will be ignored if you do).

Oh, I guess we already have a Postgres container and set the DATABASE_URL as an env var in our web container, so the Docker setup doesn't need that set in .env. I'll update it

dev-guide.md
erwa: Hmm, in later sections, it already says to contact a collaborator for env settings like keys or to just get them from https://dashboard.heroku.com/apps/smallboard/settings

I'd rather not put a specific person since that might fall out-of-date over time

erwa:
Updated the docs and .env.test. @rgossiaux , @asdfryan , let me know if you have other comments

rgossiaux (approved):
Thanks!

PR #314, opened 2021-01-08
Fix node_modules Docker issues

### LINKED ISSUE #313: webpack not found when running docker-compose up in fresh checkout
In a fresh checkout of smallboard, after install prereqs like Docker and doing `cp .env.test .env`, if you run `docker-compose up`, there'll be an error in the `web` container log

```
web_1  | yarn run v1.22.4
web_1  | $ webpack --config webpack.dev.js ./hunts/src/index.js --output-path ./hunts/static
web_1  | /bin/sh: webpack: not found
web_1  | error Command failed with exit code 127.
```

If you open Small Board in a browser and try to open a hunt, it'll error out because `main.js` won't be found:
```
web_1  | [07/Jan/2021 22:40:53] "GET /static/main.js HTTP/1.1" 404 1647
```

Looking inside the container, it seems the entire `node_modules` folder is missing.

Manually running `yarn install` inside the container creates the `node_modules` folder and then restarting the container fixes the issue, but ideally the `Dockerfile` that creates the `web` container's image would be updated.

I encountered this while running Small Board in WSL2 on Windows.

rgossiaux:
Fixes #313.

The node_modules/ directory generated in the image was getting bypassed
when we mounted the source code directory. Specifying it as another
volume somehow seems to fix this.

Also deleting the code copying line from the Dockerfile since it's
misleading and unnecessary with the volume.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,3 @@ RUN yarn install
 # Install patches
 COPY ./patches ./patches
 RUN yarn install
-
-COPY . .
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -14,6 +14,7 @@ services:
     command: python manage.py runserver 0.0.0.0:8000
     volumes:
       - .:/usr/src/smallboard/:Z
+      - /usr/src/smallboard/node_modules
     ports:
       - 8000:8000
     environment:


bcai2 (commented):
A fresh --no-cache rebuild seems to be working for me, with the listed version of webpack as well (5.9.0 as opposed to 5.11.1, which I was getting from manually downloading). LGTM!

bcai2 (approved):

erwa (approved):
Ah, so the mounted `.:/usr/src/smallboard` volume was what was overwriting the `node_modules` directory. Thanks for investigating and for the fix

PR #315, opened 2021-01-08
 Add checkbox that filters out solved puzzles

pcvera:


rgossiaux (commented):
Nice!

Poking around with this I got this strange behavior. Before: 
image
After:
image

which doesn't seem great since it makes the bottom two puzzles look like they belong to the wrong meta. 

Not actually sure what the right behavior is... polling in Discord now for some thoughts. In this case I wonder if we should show any puzzle that has any unsolved descendants, so that they have enough context. Or we could just hide the unsolved feeders since they don't matter. 

Maybe the best thing is to give the option to do either rather than guess what people want? Another checkbox or something maybe to toggle between the two?

pcvera:
I updated to have the `Priorty` and `Unsolved` filters, I did include any attempt at automatically tagging low-priority as discussed in discord, I think it wouldn't fit in the scope of this PR.

# All Puzzles
image

# Priority
image

# Unsolved
image

asdfryan:
> I updated to have the `Priorty` and `Unsolved` filters, I did include any attempt at automatically tagging low-priority as discussed in discord, I think it wouldn't fit in the scope of this PR.
> 

Nice! I filed https://github.com/cardinalitypuzzles/smallboard/issues/322 for the auto-tagging feature.

This looks great! I'm guessing that "Unsolved" filters for unsolved puzzles and their (potentially solved) metas, whereas "Priority" filters for all unsolved, non-lo-pri puzzles.

I wonder if it's possible to make that a little more clear from the name... not sure.

hunts/src/HuntViewMain.js
```diff
@@ -147,7 +152,50 @@ export const HuntViewMain = (props) => {
...
+              type="radio"
+              checked={filterSolved === SOLVE_STATE_FILTER_OPTIONS.UNSOLVED}
+              onChange={(evt) => {
+                if (evt.target.checked) {
+                  setFilterSolved(SOLVE_STATE_FILTER_OPTIONS.UNSOLVED);
+                }
+              }}
+            ></input>
+            Unsolved
+          </label>
```
rgossiaux, line 196: Could consider making this more succinct with a .map over some suitable array

hunts/src/solveStateFilter.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function isSolved(tableRow) {
+  return !isUnsolved(tableRow);
+}
+
+function hasUnsolvedChildren(tableRow) {
+  return tableRow.subRows.filter(isUnsolved).length !== 0;
+}
+
```
rgossiaux, line 18: I think we need to do this recursively / traverse all the descendants; we care about if any descendant is unsolved to give the full context.

hunts/src/solveStateFilter.js
```diff
@@ -0,0 +1,47 @@
...
+  if (filterValue === SOLVE_STATE_FILTER_OPTIONS.PRIORITY) {
+    const rowIdsToExclude = new Set();
+    rows.forEach((row) => {
+      if (isSolved(row)) {
+        // Explicitly exclude this row.
+        rowIdsToExclude.add(row.id);
+        // Also exclude its children.
+        row.subRows.forEach((subRow) => {
+          rowIdsToExclude.add(subRow.id);
+        });
```
rgossiaux, line 33: Same here. Maybe the easiest would be to have a function to get all the descendants of a row & then you could just iterate over that in both places.

rgossiaux (changes requested):
Cool, I like the radio boxes. This is going to make it a lot easier to handle figuring out what to work on later in the hunt when we have lots of puzzles solved.

hunts/src/solveStateFilter.js
pcvera: good catch, fixed

rgossiaux (approved):
Cool, thanks!

PR #316, opened 2021-01-08
Remove /hunts/ from some API endpoints

rgossiaux:
The nesting was too deep and unnecessary for puzzle tags and puzzle
answers.

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -55,20 +55,18 @@ urlpatterns = [
         "v1/hunts/<int:hunt_id>/puzzles/<int:pk>", puzzle_detail, name="puzzle_detail"
     ),
     path(
-        "v1/hunts/<int:hunt_id>/puzzles/<int:puzzle_id>/answers",
+        "v1/puzzles/<int:puzzle_id>/answers",
         answer_list,
         name="answer_list",
     ),
     path(
-        "v1/hunts/<int:hunt_id>/puzzles/<int:puzzle_id>/answers/<int:pk>",
+        "v1/puzzles/<int:puzzle_id>/answers/<int:pk>",
         answer_detail,
         name="answer_detail",
     ),
+    path("v1/puzzles/<int:puzzle_id>/tags", tag_list, name="tag_list"),
     path(
-        "v1/hunts/<int:hunt_id>/puzzles/<int:puzzle_id>/tags", tag_list, name="tag_list"
-    ),
-    path(
-        "v1/hunts/<int:hunt_id>/puzzles/<int:puzzle_id>/tags/<int:pk>",
+        "v1/puzzles/<int:puzzle_id>/tags/<int:pk>",
         tag_detail,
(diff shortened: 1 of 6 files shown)


api/views.py
```diff
@@ -206,8 +205,7 @@ class PuzzleTagViewSet(viewsets.ModelViewSet):
     serializer_class = PuzzleTagSerializer
 
     def get_queryset(self):
-        hunt_id = self.kwargs["hunt_id"]
-        return PuzzleTag.objects.filter(hunt__id=hunt_id)
+        return PuzzleTag.objects.all()
```
asdfryan, line 208: Hmm this is a real behavior difference right? Do we imagine hunt-specific tags to be a thing? If not, we should remove the `hunt` field from PuzzleTags.

api/urls.py
```diff
@@ -55,20 +55,18 @@
         "v1/hunts/<int:hunt_id>/puzzles/<int:pk>", puzzle_detail, name="puzzle_detail"
     ),
     path(
-        "v1/hunts/<int:hunt_id>/puzzles/<int:puzzle_id>/answers",
+        "v1/puzzles/<int:puzzle_id>/answers",
```
asdfryan, line 58: Can we make `v1` a variable and share it across `urls.py` and test files?

rgossiaux: I'm a little hesitant because I don't want it to ever vary (we should never change it--the only reason to bump the version would be to maintain backwards compatibility => the existing code & tests should continue using the same endpoints) and it feels like any constant is going to be less readable than just the two characters "v1". So I bias towards just leaving it until it's actually a problem

api/views.py
rgossiaux: There shouldn't be a behavior difference. This queryset is just the starting point that gets further filtered later (`destroy` filters it by `pk` in `self.get_object()` I think). With that said maybe I should make it puzzle.tags.all() to more accurately reflect what this class serves (tags for a specific puzzle).

As for `hunt` we definitely want tags across different hunts to be independent from each other so that's why this is here. It's technically duplicative since right now we assume that any tag is associated with at least one puzzle & you could get the hunt from that relationship, but eventually I was hoping to remove that constraint I think, for storing default tags.

asdfryan: I see -- thanks!

asdfryan (approved):

api/views.py
rgossiaux: (I updated this to `PuzzleTag.objects.filter(puzzles__id=puzzle_id)` which is more precise)

hunts/src/api.js
```diff
@@ -1,5 +1,7 @@
 import Cookies from "js-cookie";
 
+const API_PREFIX = "/api/v1";
```
erwa, line 3: can we reference this constant from the tests as well?

erwa (approved):
Looks good!

PR #317, opened 2021-01-08
Fix modals causing table rerenders

### LINKED ISSUE #290: React UI completely re-rendering when opening/closing modals
Noticed that there's a substantial (several hundred ms) lag when I open/close modals using a copy of last year's mystery hunt. Haven't dug into it but I assume the whole table is rerendering when this happens. 

I haven't really spent time looking focusing on the performance of the React UI so there are probably a lot of other performance issues like this.

rgossiaux:
Fixes #290

Opening/closing a modal re-renders the parent component; since the PuzzleTable
is a pure component we can wrap in React.memo and only rerender when
props (ie data) change.

### DIFF
--- a/hunts/src/puzzle-table.js
+++ b/hunts/src/puzzle-table.js
@@ -38,7 +38,7 @@ function rowClassName(row) {
   }
 }
 
-export function PuzzleTable({ columns, data, filter }) {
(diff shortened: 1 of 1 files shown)


asdfryan (approved):
Thanks!

PR #318, opened 2021-01-08
Basic stats + Basic stats page

bcai2:
* Added basic stats in `hunts/stats_utils.py`, containing functions for counting solved, unsolved, and unlocked puzzles for a hunt as well as meta solves.
* Added a unit test for the basic hunt stats (along with a setup for unit testing hunts).
* Added a basic stats page at `<hunt_name>/stats`. Currently, it displays the four stats mentioned above on each (manual) refresh. (Count numbers are bolded for now. I'm currently split between whether bolded or un-bolded numbers look better.)

image

* Added a hyperlink (to the right of the hunt name in the React setup) from the hunt page to the stats page and another link in the other direction.

image


hunts/tests.py
```diff
@@ -1,3 +1,86 @@
...
+from accounts.models import Puzzler
+from .models import Hunt
+from puzzles.models import Puzzle
+from answers.models import Answer
+from .stats_utils import *
+
+from django.utils import timezone
+from datetime import timedelta
+
 # Create your tests here.
```
asdfryan, line 12: Remove?

asdfryan (commented):
Nice!

Personally, I think it would also be nice to display the 4 main stats (minus the graph) in the main puzzle table page as well (right above the search bar).

bcai2:
Done!
image
These update with the 10-second puzzle refresh as well.

hunts/src/HuntViewMain.js
```diff
@@ -138,7 +145,41 @@ export const HuntViewMain = (props) => {
...
+            {hunt.num_solved}
+          </Col>
+          <Col xs={1} className="text-danger">
+            {hunt.num_unsolved}
+          </Col>
+          <Col xs={1} className="text-secondary">
+            {hunt.num_unlocked}
+          </Col>
+        </Row>
+      </Container>
```
rgossiaux, line 182: Maybe we can split this out into another component, this file/component is already kinda busy (my fault)

api/serializers.py
```diff
@@ -11,9 +11,35 @@
...
+        fields = (
+            "id",
+            "name",
+            "active",
+            "url",
+            "num_solved",
+            "num_unsolved",
+            "num_unlocked",
+            "num_metas_solved",
+        )
```
rgossiaux, line 42: What do you think about computing this on the front-end instead? I think I have an aesthetic preference for that vs putting all this derived data inside every Hunt object but I think it also might have a couple other advantages (always in sync with the other data you have & no need for another set of recurring API calls).

In this case I think it would work nicely as a bunch of redux selectors in `puzzlesSlice.js`

rgossiaux (commented):
Nice, thank you for implementing this!!

api/serializers.py
bcai2: Is there a nice way to link this up with the stats page if we do this on the front-end? Ideally, I'm aiming for this to also be displayed there, along with being polled (or at least updated) regularly with the progress chart and the other stats.

bcai2: Alternatively, we could keep the back-end functions for the stats page and move the puzzle/hunt display page's computation to a front-end one. (This does do the same thing two different ways; not sure if that's a problem...)

rgossiaux: I think what you have right now should be fine for the stats page, right? I think having the Hunt.num_foo() functions is fine, just not sure about the serialization alongside every other hunt. If you want endpoints that you can query for refreshing the standalone stats page (which seems reasonable--wasn't sure if you were planning on adding that) maybe something like a /hunts/:id/stats endpoint that just returns them directly?

I do have some preference for deriving them in the front-end for the main puzzles page so that they're always in sync with the data we've got.

bcai2: Sounds good. I'll look into it. (Refreshing the page is definitely part of the eventual goal, part of the next branch I was looking into; hopefully I'll be able to get there with enough time to test and spare.)

bcai2:
Time-based stats have been added to the stats page as a table below the 4 main stats (and eventually, the progress graph). If no start time is given or if the current time isn't after the start time, these are ignored (an "N/A" is returned). The time elapsed is computed as `min(current_time, end_time [if given]) - start_time`. 

image

A stat for progression (total solves + # of unsolved puzzles with solved metas) might come in handy as well, but I'll probably add that later (if lo-pri tags are implemented).

(The back-end stats computations on the main (puzzle) page have been replaced with front-end computations as well.)

hunts/src/HuntViewHeader.js
```diff
@@ -0,0 +1,54 @@
...
+import Container from "react-bootstrap/Container";
+import Row from "react-bootstrap/Row";
+import Col from "react-bootstrap/Col";
+
+function HuntViewHeader({
+  hunt,
+  numMetasSolved,
+  numSolved,
+  numUnsolved,
+  numUnlocked,
```
rgossiaux, line 11: nit: selecting these inside this component might be slightly cleaner

hunts/src/HuntViewMain.js
```diff
@@ -1,13 +1,21 @@
...
 import PropTypes from "prop-types";
 import { useSelector, useDispatch } from "react-redux";
-import { fetchPuzzles, selectPuzzleTableData } from "./puzzlesSlice";
+import {
+  fetchPuzzles,
+  selectPuzzleTableData,
+  getNumUnlocked,
+  getNumSolved,
+  getNumUnsolved,
+  getNumMetasSolved,
```
rgossiaux, line 10: nit: our other selectors are prefixed `select` instead of `get` so I might rename them that

rgossiaux (approved):
Nice, thanks! A couple small comments inline. Thanks for taking this on!

asdfryan (approved):

PR #319, opened 2021-01-08
_get_service_choices() now lists options directly

leonyz:


chat/models.py
```diff
@@ -8,7 +8,7 @@ def _get_default_service():
 
 
 def _get_service_choices():
-    return [(p, p) for p in settings.CHAT_SERVICES.keys()]
+    return [("DISCORD", "DISCORD")]
```
asdfryan, line 11: Now that this is static and no longer depends on env vars, I actually think we should just remove this function and keep these as class variables. See STATUS_CHOICES as an example inside answers/models.py.

asdfryan (changes requested):
Thanks! Just one nit.

asdfryan (approved):

Issue #169, opened 2020-01-21
Scripts broken in cloned template spreadsheet


erwa:
When the template Google spreadsheet was cloned for each puzzle, the JavaScript functions (as illustrated in the "Examples" tab) did not work for some reason. If you manually copy the spreadsheet, the JavaScript does work.

rgossiaux:
I feel like someone fixed this already a long while ago but if not we should definitely address this before hunt this year.

rawxfish:
sooooooooooo this is actually broken. 

![image](https://user-images.githubusercontent.com/1312469/104084828-c4d63f00-5218-11eb-96f3-e3494a84c0cb.png)

PR #323, opened 2021-01-09
transfer ownership of sheets after creation

rawxfish:
The scripts were broken (#169) because copying the sheets template sets the owner to a service account, which is unable to run scripts. This change transfers ownership back to original owner of template.

### DIFF
--- a/google_api_lib/task.py
+++ b/google_api_lib/task.py
@@ -41,18 +41,6 @@ def create_google_sheets_helper(self, name):
     return file
 
 
-def transfer_ownership(self, file_id):
-    response = (
-        self._drive_service.permissions()
-        .create(
-            fileId=file_id,
-            body=req_body,
-            fields="webViewLink",
-        )
-        .execute()
-    )
-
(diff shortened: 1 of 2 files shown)


google_api_lib/task.py
```diff
@@ -15,16 +15,42 @@
...
+    self._drive_service.permissions().create(
+        fileId=file["id"],
+        body=transfer_permission,
+        transferOwnership=True,
+    ).execute()
+
+    return file
+
+
+def transfer_ownership(self, file_id):
```
erwa, line 44: This looks unused? (I think the logic is already in create_google_sheets_helper)

erwa (commented):
Ah, nice find on why the scripts didn't work when copied via script. Looks good overall, one comment.

google_api_lib/task.py
rawxfish: whoops thought I deleted that

erwa (approved):

PR #324, opened 2021-01-09
Improve performance of main /puzzles API call

rgossiaux:
I installed django debug toolbar locally and used it to profile the API.
I found that on last year's hunt we were making 532 queries to load all
the puzzles due to bad prefetching:

* We were prefetching metas but not feeders
* We were sorting the tags which invalidated the prefetch for them
* We were filtering guesses to only show correct ones which invalidated
  the prefetch for them as well.

To fix, this adds prefetching for feeders, moves the tag sorting into
the tag model, and does the correct answer filtering inside the
prefetch. This brings it down to only 7 queries for last year's hunt
which is much more reasonable.


asdfryan (approved):
Thanks! Amazing PR 👍

erwa:
Very nice. Just curious, what is the puzzles page load time before and after this change, as measured from the developer tools in your browser?

rgossiaux:
e2e time on my computer is pretty variable depending on what else is going on. On my copy of last year's hunt locally (+ with profiling on and some dev/debug stuff installed), on master it's usually some amount over 2s to fetch all the puzzles, and then around 1.5s to render them after that. With this change fetching the puzzles now takes around 700ms or so. I wouldn't take the absolute numbers super seriously though, we should profile prod for that.

PR #325, opened 2021-01-09
Use priority tags when sorting puzzles

leonyz:
Puzzles are sorted by priority tag (high priority > no tag > low priority). Sort is within puzzle type, so low-priority feeders still appear above all metas.


hunts/src/puzzlesSlice.js
```diff
@@ -96,6 +96,30 @@ function puzzleComparator(a, b) {
   } else if (a.is_meta && !b.is_meta) {
     return 1;
   }
+  // High-priority first
```
asdfryan, line 99: If it were me, I would:

1) Write a helper function `priority(row)` that returns 1 if hi-pri, else if-1 if lo-pri, else 0.
2) Return `priority(a) - priority(b)` if nonzero. Else, fall back to b.id - a.id;

leonyz: Latest commit follows your suggestion.

hunts/src/puzzlesSlice.js
```diff
@@ -83,6 +83,15 @@ export const addPuzzleTag = createAsyncThunk(
   }
 );
 
+function priority(row) {
```
asdfryan, line 86: Can you scope this inside puzzleComparator()?

Also, for consistency could you you return 1 for Hi-pri and -1 for low-pri, and return `priority(b)-priority(a)`?

asdfryan (approved):

rgossiaux:
Thank you!!

Issue #327, opened 2021-01-09
Handle puzzle creation errors


rawxfish:
#323 transfers ownership from service account to real account, but unfortunately this is a rate limited operation and sheet creation sometimes fails...and does not retry. 

```
googleapiclient.errors.HttpError: <HttpError 403 when requesting https://www.googleapis.com/drive/v3/files/1-Lxra3kdFd5nLtDOJPdhZQql2Zx5Nq9JAaYhswqJXIg/permissions?transferOwnership=true&alt=json returned "Rate limit exceeded. User message: "Sorry, you have exceeded your sharing quota."". Details: "Rate limit exceeded. User message: "Sorry, you have exceeded your sharing quota."">
```

rgossiaux:
Separate from the root cause w/ Google, are we retrying our celery tasks on failure?

rawxfish:
We are now (up to five times)!

PR #329, opened 2021-01-09
Fix sheets 

rawxfish:
A few changes for #327 that hopefully makes sheets creation more robust:

- Separates sheets creation and transferring of sheets ownership from service account to user account. The latter was failing due to a sharingRateLimitExceeded error, which according to [this stackoverflow](https://stackoverflow.com/questions/46844246/drive-api-sharingratelimitexceeded-error) is because there is a 50 permission creation per day, although I couldn't find this quota in the official google API docs. 
- Switch from permission creation to permission update, which for some reason is not quota-ed. I wrote a script to update the permissions for the hundreds of existing sheets and I didn't see that error again, so hopefully that means this will not break mid-hunt. Even if it does break, creation and ownership are separate now, so we'll at least have sheets even if the custom functions won't work, which is not the best but not so bad.
- Add auto retries with backoff

### DIFF
--- a/google_api_lib/utils.py
+++ b/google_api_lib/utils.py
@@ -1,8 +1,8 @@
+from django.conf import settings
 import googleapiclient
 import googleapiclient.discovery
-from django.conf import settings
+import googleapiclient.errors
 from google.oauth2 import service_account
-
 
 from celery import Task
 
@@ -12,6 +12,12 @@ def enabled():
 
 
 class GoogleApiClientTask(Task):
+    autoretry_for = (googleapiclient.errors.Error,)
+    retry_kwargs = {"max_retries": 5}
+    retry_backoff = True
+    retry_backoff_max = 600
+    retry_jitter = True
+
     def __init__(self):
         if not enabled():
             return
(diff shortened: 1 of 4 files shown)


rawxfish:
Not directly, but somewhat related: currently we just have a single worker for all the celery google related tasks. This is so I could avoid thinking about what would happen if multiple workers edited a sheet. But it'd be sorta bad if our poor single worker died midhunt (idk how likely this is)

I'm thinking of adding a celery queue for sheet creation that has many workers, and a separate queue for any operation that changes the sheets (title changes, updating feeder answers in the meta sheets) with a single worker. I imagine sheet creation is the most important thing anyways.

rgossiaux:
> Not directly, but somewhat related: currently we just have a single worker for all the celery google related tasks. This is so I could avoid thinking about what would happen if multiple workers edited a sheet. But it'd be sorta bad if our poor single worker died midhunt (idk how likely this is)
> 
> I'm thinking of adding a celery queue for sheet creation that has many workers, and a separate queue for any operation that changes the sheets (title changes, updating feeder answers in the meta sheets) with a single worker. I imagine sheet creation is the most important thing anyways.

I thought (not sure) that there's a limit of 1 celery worker in the free tier, is that true? Besides cost this seems good.

We should probably try intentionally crashing the celery worker on Heroku and see what happens, lol. Hopefully it autoheals + the Coralogix integration makes a Discord alert for it

rgossiaux (approved):
thank you!!

google_api_lib/utils.py
```diff
@@ -20,16 +26,21 @@ def __init__(self):
...
             .execute()
         )
         self._sheets_owner = template["owners"][0]["emailAddress"]
+
+    def drive_service(self):
+        return googleapiclient.discovery.build(
+            "drive", "v3", credentials=self._credentials, cache_discovery=False
+        )
+
+    def sheets_service(self):
```
erwa, line 43: I think calling sheets_service() and drive_service() will cause a new Google API client to be constructed each time the method is called. I had these initialized once in the constructor and stored as _sheets_service and _drive_service to avoid this overhead for every use. I think what we had before is preferable?

rawxfish: The old code doesn't work with celery retries. I think related to some thread safety issue: https://stackoverflow.com/questions/33557857/google-datastore-ssl-errors-in-python-from-google-compute-instances

erwa: Hmm, I see. Seems there are a lot of googleapis issues and pull requests related to similar issues. I guess this is ok for now to workaround the issue. We can revisit if we notice performance issues related to sheets operations.

Issue #326, opened 2021-01-09
Add puzzles modal does not close even after puzzle is created in the table


asdfryan:
See screenshot below: 
![Screen Shot 2021-01-09 at 11 26 18 AM](https://user-images.githubusercontent.com/1498449/104106914-b9f9c980-526d-11eb-923e-859c3b1a3915.png)

You can see that the puzzle has already been created (in the table in the background).

rgossiaux:
Anything in your browser console?

asdfryan:
Update: This only reproduces if `REDIS_URL` is not set in your `.env`.

rgossiaux:
After #330 it will close eventually, at least, though I think we can still improve. Copying from there:

>Two different UX options:
>
>* grey out the modal while the add puzzle request is in progress
>* close the modal immediately when you make the add puzzle request
>
>I was worried that closing the modal immediately could be misleading sometimes but in this case it seems better than just blocking the screen lol

PR #330, opened 2021-01-10
Set max retries on Celery connecting to Redis

rgossiaux:
For #326

This still leaves the modal open for a number of seconds while we retry
connecting to redis so it's only a bandaid.


rawxfish (approved):

rgossiaux:
Two different UX options:

* grey out the modal while the add puzzle request is in progress
* close the modal immediately when you make the add puzzle request

I was worried that closing the modal immediately could be misleading sometimes but in this case it seems better than just blocking the screen lol

asdfryan (approved):

PR #331, opened 2021-01-10
Disable chat features in api unit tests

asdfryan:
Ignore whatever chat settings in your local .env file.

### DIFF
--- a/api/tests.py
+++ b/api/tests.py
@@ -1,10 +1,16 @@
 from .test_helpers import SmallboardTestCase
+from django.test import override_settings
 from rest_framework import status
 from puzzles.models import Puzzle, PuzzleTag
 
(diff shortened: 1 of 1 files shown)


akirabaruah (approved):
Seems like a nice, simple way to unblock the tests. Do we intend to mock/fake chat features for these tests eventually? I could see a case being made either way, since we have discord-specific testing.

asdfryan:
> Seems like a nice, simple way to unblock the tests. Do we intend to mock/fake chat features for these tests eventually? I could see a case being made either way, since we have discord-specific testing.

IMO It's not necessary. This code is 2 levels removed from the ChatService API. discord_lib has its own unit tests, and chat/ has unit tests that fake the service.

PR #332, opened 2021-01-10
Auto-create new PUZZLES / ARCHIVE categories when they become full, and put new channels there instead

asdfryan:


discord_lib/discord_chat_service.py
```diff
@@ -46,11 +48,36 @@ def delete_channel(self, channel_id):
         if int(channel_id) in channels_by_id:
             self._client.channels_delete(int(channel_id))
 
+    def get_or_create_category(self, category_name):
+        """
+        Returns category that has fewer than 50 children channels. If none
```
akirabaruah, line 53: nit: consider replacing "50 children channels" with "max number of children channels specified in the constructor" or something like that, since it is parameterized

akirabaruah (approved):
Loving the tests! Looks great overall. I'm wondering if it would get tedious for folks to search among multiple folders to find a puzzle chat. Perhaps beyond the scope of this PR, maybe we should have some sort of index? Or maybe just have a bot command to query?

discord_lib/discord_chat_service.py
asdfryan: Done.

asdfryan:
> Loving the tests! Looks great overall. I'm wondering if it would get tedious for folks to search among multiple folders to find a puzzle chat. Perhaps beyond the scope of this PR, maybe we should have some sort of index? Or maybe just have a bot command to query?

This is a good point -- although I suspect this may only be applicable for `ARCHIVE`. I suspect we won't have 25+ open puzzles at a time.

akirabaruah:
Good point. In practice, we'd probably have to look at 1 or 2 active puzzle folders max during MITMH.

PR #333, opened 2021-01-10
more workers, more queues

rawxfish:
As mentioned in #329:

> Not directly, but somewhat related: currently we just have a single worker for all the celery google related tasks. This is so I could avoid thinking about what would happen if multiple workers edited a sheet. But it'd be sorta bad if our poor single worker died midhunt (idk how likely this is)
>
> I'm thinking of adding a celery queue for sheet creation that has many workers, and a separate queue for any operation that changes the sheets (title changes, updating feeder answers in the meta sheets) with a single worker. I imagine sheet creation is the most important thing anyways.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,8 @@
 release: python manage.py migrate --noinput
 web: gunicorn smallboard.wsgi --log-file -
+
+cgs_worker1: celery -A smallboard worker -l INFO -Q cgs_queue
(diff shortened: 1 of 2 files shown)


rawxfish:
Draft b/c not sure if this will give us more free workers, testing on review app

rawxfish:
It does not. Abandon ship.

PR #335, opened 2021-01-10
Remove voice chat link (currently useless), and stop calling Discord API to generate invite links

asdfryan:
Stop generating invite links and use /guild_id/category_id instead. Also stop showing voice links which are useless.


akirabaruah (approved):
LGTM!

rgossiaux (approved):
Thanks!

PR #337, opened 2021-01-10
Rename sheet when puzzle is solved

### LINKED ISSUE #135: Rename puzzle spreadsheet as [SOLVED: THEANSWER] ... when puzzle is solved
Might be a useful way to notify people working on the puzzle that it's been solved, in case they missed the Slack message and didn't notice on the site and nobody told them.

rgossiaux:
Fixes #135

Didn't put the answer in there since multiple answers makes it a little
more complicated; figure this is the important info anyway.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -71,6 +71,12 @@ class AnswerViewSet(viewsets.ModelViewSet):
                 transaction.on_commit(
                     lambda: AnswerViewSet.__update_meta_sheets_for_feeder(puzzle)
                 )
+                if google_api_lib.enabled() and puzzle.sheet:
+                    transaction.on_commit(
+                        lambda: google_api_lib.task.rename_sheet.delay(
+                            sheet_url=puzzle.sheet, name=f"[SOLVED] {puzzle.name}"
+                        )
+                    )
             puzzle.save()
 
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -71,6 +71,12 @@ def create(self, request, **kwargs):
                 transaction.on_commit(
                     lambda: AnswerViewSet.__update_meta_sheets_for_feeder(puzzle)
```
asdfryan, line 72: Come to think of it -- this line should also be wrapped by `if google_api_lib.enabled()`.

rgossiaux: Aha hmm-- `update_meta_sheets_for_feeder` used to have a check for this inside and it looks like as part of #295 @rawxfish moved the check inside the GoogleApiClientTask __init__ in google_api_lib/utils.py.

rgossiaux: Testing locally I can confirm that the celery worker complains so I'll just fix on this branch.

rgossiaux: Split it off for clarity: #338

rawxfish: Whoops sorry about that

rawxfish (approved):

Issue #68, opened 2019-12-15
Meta Assignment in "Add Puzzle" dialog


maximized:
Minor: Sometimes the meta is known at creation time.

rgossiaux:
I'll add this when I implement this modal in the react rewrite (#251 )

PR #339, opened 2021-01-10
Try revamping the edit tags modal

rgossiaux:
Instead of a display of tags at the top that you can click to add, and a
list of all your tags at the bottom that you can use to remove, try just
toggling each tag on/off.

Played around a little while with the visuals and this is the best I had
so far but it's still a little ugly imo.

Thinking of using a UI like this to implement #68


rgossiaux:
Looks something like this:

image

Feel free to play around locally or with the review app, I wanna know what people think / how it feels.

pcvera (approved):

rgossiaux:
Yolo pushing this so we can test it but I am more than open to making changes

Issue #218, opened 2020-11-29
Clickable Puzzle tags


akirabaruah:
Puzzles have tags. Users should be able to click on tags in the UI to filter the Puzzle list to include Puzzles with the selected tag only.

rgossiaux:
Should wait for the react rewrite (#251 ) before adding this

rgossiaux:
This is unblocked now.

To do this, I think we will want to use the `useFilters` hook on react-table to add a column filter for the tags column.

Unfortunately I think the default behavior for this is likely the same bad behavior there was for global filters: a subrow only appears if all of its ancestors match the filter (because internally react-table treats subrows as a top-down tree traversal, and any node that doesn't match gets pruned, which prunes all its descendants). 

To fix this we will probably need to patch useFilters.js in react-table like we patched useGlobalFilters.js. You can look at the existing patch file to see what that looked like (for filters it was a relatively small change though the logic isn't totally obvious). 

In order to generate the patch, you will need to:

* Navigate to the react-table directory in your node_modules
* Install the dependencies (`yarn install`) of react-table locally.
* Make your changes to the useFilters.js file
* Run `yarn run build` to rebuild the react-table.js output file with your changes
* Remove the new files you added from `yarn install` (I think `react-table/yarn.lock` as well as the `react-table/node_modules` subdirectory) so they don't get included in the patch.
* Go back to the smallboard root and run `yarn patch-package react-table` which should generate a new patchfile for the react-table library.
* Inspect the patch and make sure nothing crazy happened, and if it looks good, commit it!

Those steps are from memory so I may have missed something but I believe that's approximately the workflow.

Besides patching react-table, I think you can follow https://react-table.tanstack.com/docs/api/useFilters to see how to add a filter (there is example source linked there which should be helpful).

It might be easiest to use redux to keep track of the filter state, you can figure it out when implementing.

This is related to #220 in that I think that one is essentially a different column filter (filtering on the answers column instead). In particular I think they both have a dependency on the react-table patch changes mentioned here, and beyond that the implementations will probably look somewhat similar (though this one has a more complicated UI).

rgossiaux:
@dan-simon if you're still interested in this issue feel free to grab the assignment. I'm not sure I did a good job conveying the scope on Discord so if you think you won't have time that's okay too.

rgossiaux:
Merged in #404

PR #343, opened 2021-01-10
Add clicking on tags

dan-simon:
This code is a mess and I'm sure I'm missing something. There might be some dead code from previous attempts I made to do this with a different approach.

Fixes https://github.com/cardinalitypuzzles/smallboard/issues/218


hunts/src/tagFilter.js
```diff
@@ -0,0 +1,24 @@
...
+  const allChildren = [tableRow];
+  for (let i = 0; i < allChildren.length; i++) {
+    const current = allChildren[i];
+    Array.prototype.push.apply(allChildren, current.subRows);
+  }
+  return allChildren;
+}
+
+export function addOrRemoveTag(tagList, tag) {
+  let tagInList = tagList.filter(x => ['name', 'color', 'id'].every(key => x[key] === tag[key]))[0];
```
akirabaruah, line 11: Can we use `tagList.find(...)` instead of `tagList.filter(...)[0]` here?

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,24 @@
+function getDescendants(tableRow) {
+  const allChildren = [tableRow];
+  for (let i = 0; i < allChildren.length; i++) {
+    const current = allChildren[i];
+    Array.prototype.push.apply(allChildren, current.subRows);
+  }
+  return allChildren;
+}
+
+export function addOrRemoveTag(tagList, tag) {
```
akirabaruah, line 10: Can we leave a comment for this func explaining that it essentially toggles the tag's presence in the array? It wasn't super clear to me from the func name, so even something like "Adds the tag if it doesn't exist or removes it if it does" would be great.

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,24 @@
...
+  let tagInList = tagList.filter(x => ['name', 'color', 'id'].every(key => x[key] === tag[key]))[0];
+  return (tagInList !== undefined) ? tagList.filter((i) => i !== tagInList) : tagList.concat(tag);
+}
+
+export function filterPuzzlesByTagfn(rows, id, filterValue) {
+  let requiredRowIDs = rows.filter((row) => {
+    return tagList.length === 0 || tagList.some(tag => row.original.tags.map(x => x.id).includes(tag.id));
+  }).map(row => row.id);
+  // Show metas with puzzles with tags we want? Not sure if this is a good idea.
```
akirabaruah, line 19: I'd advocate for keeping it simple and only match direct tags, not descendants, since I imagine users might get confused to see puzzles listed that don't directly match tags. I could be convinced otherwise though :D

hunts/src/HuntViewMain.js
```diff
@@ -195,6 +202,19 @@ export const HuntViewMain = (props) => {
...
+        <div>
+          {filterTags.length > 0 ? 'Viewing puzzles with tags: ' : ''}
+          {filterTags.map(({ name, color, id }) => {
+            return <TagPill
+              name={name}
+              color={color}
+              id={id}
+              puzzleId={null}
+              key={name}
+              onDelete={function () {addOrRemoveFilterTag({ name, color, id })}}
```
akirabaruah, line 214: Do we intend to use only the "remove" part of `addOrRemoveFilterTag` here?

akirabaruah (changes requested):
LGTM overall with some minor comments, but also wondering why you didn't end up needing to patch the useFilters hook mentioned by @rgossiaux in #218.

hunts/src/HuntViewMain.js
dan-simon: Yeah, we do. I could create a separate function to do that if desired.

hunts/src/tagFilter.js
dan-simon: OK, will change that, hopefully it doesn't make things break. (I wasn't actually testing with a meta, might do that.)

hunts/src/tagFilter.js
dan-simon: Yeah, didn't know find existed. (Also I'm a bit worried about this code, it uses the little-known fact that === on objects is "are these the same object" rather than "always false" or "do these have the same key-value pairs")

dan-simon:
> LGTM overall with some minor comments, but also wondering why you didn't end up needing to patch the useFilters hook mentioned by @rgossiaux in #218.

Probably due to not testing with metas. I'll try to fix that now.

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+  return allChildren;
+}
+
+export function addOrRemoveTag(tagList, tag) {
+  let tagInList = tagList.filter(x => ['name', 'color', 'id'].every(key => x[key] === tag[key]))[0];
+  return (tagInList !== undefined) ? tagList.filter((i) => i !== tagInList) : tagList.concat(tag);
+}
+
+export function filterPuzzlesByTagfn(rows, id, filterValue) {
```
akirabaruah, line 15: nit: Should we change `Tagfn` to `TagFn` for upper camel case? The existing `filterTypes` don't seem to be consistent, but uppercase `Fn` seems slightly better?

hunts/src/puzzle-table.js
```diff
@@ -45,11 +46,12 @@ function rowClassName(row) {
 }
 
 export const PuzzleTable = React.memo(
-  ({ columns, data, filter, filterSolved }) => {
+  ({ columns, data, filter, filterSolved, filterTags, addOrRemoveFilterTag }) => {
     const filterTypes = React.useMemo(
       () => ({
         globalFilter: textFilterFn,
         solvedFilter: filterSolvedPuzzlesfn,
+        tagsFilter: filterPuzzlesByTagfn
```
akirabaruah, line 54: nit: consider adding `,` at the end of this line

hunts/src/tagFilter.js
akirabaruah: Thoughts on renaming this to `toggleTag()`?

hunts/src/tagFilter.js
akirabaruah: Doesn't `===` just check for type and value equivalence (not necessarily the same object in memory)? If so, this should be okay, yeah?

hunts/src/tagFilter.js
akirabaruah: oh yeah, might be good to sanity check with a meta to see if there's any weird behavior

hunts/src/HuntViewMain.js
akirabaruah: Yeah, it might be clearer to separate the "find matching tag" and "remove" (might just be an array pop or something) logic. Then for `addOrRemove` you can use `findMatchingTag`, I think?

hunts/src/tagFilter.js
dan-simon: No, it checks for same object in memory, which is (at least part of) why we need to do this tagInList thing rather than just indexOf and includes.

hunts/src/HuntViewMain.js
dan-simon: Not quite sure I see how to do this elegantly but I can try (might not be doing what you thought of though).

hunts/src/tagFilter.js
dan-simon: Somehow (once I got it to work in the first place, which right now is only locally) it seems to work with a meta, not sure I understand how. Will look into that before pushing, because I want to have some idea why this code is working.

dan-simon: I think I understand how it works now. The indentation is a bit weird (puzzles in metas are indented whether or not the meta is visible), I'll create an issue for that (also applies to the show all/priority/unsolved settings). But I don't think that's critical.

dan-simon:
Also, do we have any guarantees on uniqueness of tag IDs? That would clearly let us simplify the code a bit.

hunts/src/HuntViewMain.js
akirabaruah: Eh, if it ends up being more complicated, feel free to leave as is and add a comment here mentioning that we expect to remove only :+1:

dan-simon: I recently pushed a new version of the whole tagFilter.js file; the new version is longer but I think it's a lot clearer so I personally prefer it.

hunts/src/tagFilter.js
akirabaruah: Yeah, it seems like `===` checks for object equivalence for non-primitive type objects, but for numbers and strings, it just does value comparison (lemme know if I'm misinterpreting): http://www.ecma-international.org/ecma-262/5.1/#sec-11.9.6

Are the `name`, `color`, and `id` fields not just strings/numbers?

dan-simon: They are just strings and numbers, and so equality is by value rather than memory, I think I was being unclear here. I'm talking about the use of !== in the filter on the next line:
`  return (tagInList !== undefined) ? tagList.filter((i) => i !== tagInList) : tagList.concat(tag);`,
which only works because === does compare identity, rather than just always being false.

If `===` were always value identity I think I could do something like
`  return tagList.includes(tag) ? tagList.filter((i) => i !== tag) : tagList.concat(tag);`
without this whole tagInList business, which would've been simpler.

akirabaruah:
> Also, do we have any guarantees on uniqueness of tag IDs? That would clearly let us simplify the code a bit.

I'm pretty sure that the django ManyToMany relation between each Puzzle and Tags enforces uniqueness within tag sets per puzzle: https://github.com/cardinalitypuzzles/smallboard/blob/9d30c28b5b31ce34b7af2436da2cdb69dc485491/puzzles/models.py#L67

Assuming this uniqueness gets propagated by the API, then I think you can assume that tag IDs are unique within each puzzle row. Not sure if I'm fully answering your question though.

dan-simon:
I was thinking of just checking ID in tagFilter.js rather than checking id, name, and color. From your comment, it doesn't sound like I can do that though, because that would require per-hunt uniqueness.

dan-simon:
This raises another question though, which is if a tag has the same ID on all the puzzles with the tag. If not, then checking tag identity via ID/color/name identity (as I'm doing now) is broken.

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,33 @@
+function getDescendants(tableRow) {
```
rgossiaux, line 1: Is this used? Ctr+f on my phone doesn't give anything

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,33 @@
...
+  const allChildren = [tableRow];
+  for (let i = 0; i < allChildren.length; i++) {
+    const current = allChildren[i];
+    Array.prototype.push.apply(allChildren, current.subRows);
+  }
+  return allChildren;
+}
+
+function isMatchingTagInFilter(tagList, tag) {
+  return tagList.find(x => ['name', 'color', 'id'].every(key => x[key] === tag[key]));
```
rgossiaux, line 11: Name or id both uniquely identify a tag within a hunt

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function filterWithTagAdded(tagList, tag) {
```
rgossiaux, line 14: Not sure I understand the naming of this function

hunts/src/puzzle-table.js
```diff
@@ -117,6 +123,7 @@ export const PuzzleTable = React.memo(
           <tbody {...getTableBodyProps()}>
             {rows.map((row, i) => {
               prepareRow(row);
+              row.original.toggleTagInFilter = toggleTagInFilter;
```
rgossiaux, line 126: Hmm I don't understand the data flow in this PR yet but I think we should generally avoid mutating row.original slash the row objects if possible. Is this just attached so that it's available in the right places? If so I think it's better passed around as a prop or stored in redux

rgossiaux (commented):
Thanks for working on this! Sorry I didn't review this earlier, it's been a hectic day... I'm in bed right now so I will give this a closer look first chance after work tomorrow but I left a few initial impressions. 

I think @pcvera 's change to filter for unsolved puzzles #315 is a good model, I think your change should likely have a similar structure

hunts/src/tagFilter.js
rgossiaux: Ok I see, it's `filter` as a noun, I read it as a verb. I'd also just inline this (and the Removed counterpart).

hunts/src/puzzle-table.js
rgossiaux: Ok, yeah, looking at it more closely I think this is just so it's available inside `TagCell`. I think we should use redux instead; it'll be a little more boilerplate but overall a bit cleaner (this kind of thing was exactly why I added redux to this project to begin with, otherwise managing the flow of all these things through the component tree gets kinda messy).

hunts/src/HuntViewMain.js
```diff
@@ -90,8 +92,9 @@ const TABLE_COLUMNS = [
   {
     Header: "Tags/Metas",
     id: "tags",
-    accessor: (row) => row.tags.map(({ name }) => name).join(" "),
+    accessor: "tags",
```
rgossiaux, line 95: This will break searching by tags (eg you can search a meta name and see all puzzles with that meta because they match the tag). Two options are to either get the global filtering to work with the new `tags` object or to keep this as a string, add some kind of separator character(s) around/between the tag strings (eg put a pipe between tags or surround them with <>) so that the string search still works, and then parse that inside your filter.

hunts/src/HuntViewMain.js
```diff
@@ -112,6 +115,14 @@ export const HuntViewMain = (props) => {
   const [filterSolved, setFilterSolved] = React.useState(
     SOLVE_STATE_FILTER_OPTIONS.ALL
   );
+  let [filterTags, setFilterTags] = React.useState([]);
+  // Passing this closure around might be expensive? I doubt it's that bad though.
```
rgossiaux, line 119: Wouldn't worry about this.

rgossiaux (changes requested):
Ok, took a pass through. Thanks for working on this! Excited to add this in. A few comments inline.

hunts/src/tagFilter.js
dan-simon: No, it's left over from when it was used. Will delete.

hunts/src/tagFilter.js
dan-simon: Ok, will use name then.

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,33 @@
...
+export function filterWithTagToggled(tagList, tag) {
+  // Returns a version of the filter tag list with the tag added
+  // if the tag is not in the filter already, otherwise returns
+  // a version with the tag removed.
+  return isMatchingTagInFilter(tagList, tag) ? filterWithTagRemoved(tagList, tag) : filterWithTagAdded(tagList, tag);
+}
+
+export function filterPuzzlesByTagFn(rows, id, tagList) {
+  return rows.filter((row) => {
+    return tagList.length === 0 || tagList.some(tag => row.original.tags.map(x => x.id).includes(tag.id));
```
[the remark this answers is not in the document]
dan-simon: Going to do "matches all" for now, can change it if desired.

hunts/src/puzzle-table.js
dan-simon: This sounds like a good idea but I'm not sure how to do it. I think it involves doing something like in modalSlice.js?

hunts/src/HuntViewMain.js
dan-simon: This also turned out just to not do anything.

hunts/src/tagFilter.js
```diff
@@ -0,0 +1,12 @@
+export function filterWithTagToggled(tagList, tag) {
```
rgossiaux, line 1: nit: i would change the `filter` prefix here, it reads like a verb but I think it was intended as a noun

hunts/src/puzzle-table.js
rgossiaux: Yep, something like that. Probably `filterSlice` or `tagFilterSlice` or something. You'll be specifying something that has some state (initially `initialState`, for you probably an empty list) and some `reducers` (probably like `addTag` and `removeTag`) which can handle actions dispatched elsewhere and compute the new state. I think the modal example is pretty similar to what you want, though you don't have to worry about `extraReducers` I don't think. Let me know if that makes sense; it took me a while to internalize all of redux's abstraction layers from reading their documentation but hopefully the modal code is simple enough to follow.

rgossiaux:
I added the redux part-- going to make a new PR for merging this in (not sure what the best way of doing this in github is)

PR #345, opened 2021-01-10
Wrap chat operations with try/catch

asdfryan:


api/views.py
```diff
@@ -67,10 +67,13 @@ def create(self, request, **kwargs):
...
-                    msg = f"{puzzle.name} has been solved with {answer.text}!"
-                    puzzle.chat_room.get_service().announce(msg)
-                    puzzle.chat_room.send_message(msg)
+                    try:
+                        puzzle.chat_room.archive_channels()
+                        msg = f"{puzzle.name} has been solved with {answer.text}!"
+                        puzzle.chat_room.get_service().announce(msg)
+                        puzzle.chat_room.send_message(msg)
+                    except:
+                        logger.warn("Chat operations failed.")
```
akirabaruah, line 76: Can we add a test to ensure that if any of these methods raise an error, we still create objects in the db?

akirabaruah, line 76: Can we include the exception in the log warning? Otherwise, I think it might eat the error message here.

```
except Exception as e:
    logger.warn("Chat operations failed: %s", e)
````

**Review by akirabaruah — APPROVED** [review]

File: api/views.py (hunk 1 above)
Comment (rgossiaux) [inline api/views.py:76]: Could we catch the exception and log it in this handler? Would make debugging easier

File: api/views.py [hunk 2]
```diff
@@ -98,7 +101,10 @@ def destroy(self, request, pk=None, **kwargs):
             ) or (not puzzle.guesses.all() and puzzle.status == Puzzle.PENDING):
                 puzzle.status = Puzzle.SOLVING
                 if puzzle.chat_room:
-                    puzzle.chat_room.unarchive_channels()
+                    try:
+                        puzzle.chat_room.unarchive_channels()
+                    except:
+                        logger.warn("Chat operations failed.")
```
rgossiaux, line 107: ditto here

api/views.py
```diff
@@ -218,10 +224,13 @@ def create(self, request, **kwargs):
...
-                msg = f"{name} has been unlocked!"
-                chat_room.get_service().announce(msg)
-                chat_room.send_message(msg)
+                try:
+                    chat_room.create_channels()
+                    msg = f"{name} has been unlocked!"
+                    chat_room.get_service().announce(msg)
+                    chat_room.send_message(msg)
+                except:
+                    logger.warn("Chat operations failed.")
```
rgossiaux, line 233: ditto here

PR #363, opened 2021-01-10
rename 'priority' to 'unsolved' and remove 'unsolved' option

pcvera:
per feedback from playtest

### DIFF
--- a/hunts/src/HuntViewMain.js
+++ b/hunts/src/HuntViewMain.js
@@ -170,29 +170,36 @@ export const HuntViewMain = (props) => {
               style={{ margin: "0 5px 0 10px" }}
               type="radio"
               checked={
-                filterSolved === SOLVE_STATE_FILTER_OPTIONS.PRIORITY ||
-                filterSolved === SOLVE_STATE_FILTER_OPTIONS.UNSOLVED
+                filterSolved === SOLVE_STATE_FILTER_OPTIONS.UNSOLVED ||
+                filterSolved ===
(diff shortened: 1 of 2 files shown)


rgossiaux:
What if instead of removing this option we have a checkbox that only appears when you're in "UNSOLVED" that's like "Include puzzles with solved metas" or something? That way both options are still available and it's clear what everything does.

pcvera:
implemented @rgossiaux 's suggestion!

hunts/src/HuntViewMain.js
```diff
@@ -169,28 +169,36 @@ export const HuntViewMain = (props) => {
...
-            ></input>
             Unsolved
           </label>
+          {(filterSolved === SOLVE_STATE_FILTER_OPTIONS.PRIORITY ||
+            filterSolved === SOLVE_STATE_FILTER_OPTIONS.UNSOLVED) && (
+            <label>
+              <input
+                style={{ margin: "0 5px 0 10px" }}
+                type="checkbox"
+                checked={filterSolved === SOLVE_STATE_FILTER_OPTIONS.UNSOLVED}
```
asdfryan, line 190: Do you mind renaming PRIORITY->UNSOLVED and UNSOLVED->UNSOLVED_WITH_SOLVED_METAS or something equivalent? (also add comments with more details on what they mean, perhaps with an example)

asdfryan (approved):
Thank you! Minor nit on renaming the state + adding comments.

rgossiaux (approved):
Thank you!!

Issue #348, opened 2021-01-10
[ui] Margins for the Hunt page


akirabaruah:
I think we might want to add left/right margins since everything currently feels a bit cramped:

![image](https://user-images.githubusercontent.com/6751517/104137245-4a103f80-5350-11eb-8024-f6b8d0184a41.png)

rgossiaux:
Thanks @pcvera !

PR #364, opened 2021-01-11
add 30 pixel margins on the left and right of smallboard hunt UI

pcvera:
in service of https://github.com/cardinalitypuzzles/smallboard/issues/348


asdfryan (approved):

akirabaruah (approved):
Beautiful. Thanks @pcvera

rgossiaux:
Nice, looks a lot better. Maybe a slightly smaller value (15 or 20px or something) would work too?

PR #365, opened 2021-01-11
fix answer filter regression

pcvera:
Add `.Answer` to the search field, understanding that `.answer` was removed during multi-answer support. (please confirm @rgossiaux @asdfryan)

### DIFF
--- a/hunts/src/HuntViewMain.js
+++ b/hunts/src/HuntViewMain.js
@@ -46,6 +46,7 @@ const TABLE_COLUMNS = [
     Header: "Answer",
     accessor: (row) => row.guesses.map(({ text }) => text).join(" "),
     Cell: AnswerCell,
+    id: "answer",
   },
   {
     Header: "Status",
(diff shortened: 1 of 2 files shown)


pcvera:
it looks like it might just be generated off of the accessor and added to the values on the `.Answer` key because that's the name of the column

From table columns:
```
  {
    Header: "Answer",
    accessor: (row) => row.guesses.map(({ text }) => text).join(" "),
    Cell: AnswerCell,
  },
```

I'm not sure how I feel about relying on that behavior.

rgossiaux:
I think you can add an id of `answer` as one of the options for a column and that will be the accessor name, haven't confirmed though

pcvera:
> I think you can add an id of `answer` as one of the options for a column and that will be the accessor name, haven't confirmed though

Yes! This is a much nicer solution.

rgossiaux (approved):
lol perfect, thanks!

asdfryan (approved):
Thanks!

PR #369, opened 2021-01-11
Try adding a connection limit from Celery to Redis

rgossiaux:
Our Redis instance has a connection limit of 20 and I've noticed hitting
it (might be related to the monitoring I added too). We're barely doing
any tasks so two connections should be more than enough.

Didn't really test this locally.


rawxfish (approved):

rawxfish:
Great, let's test in prod.

Issue #353, opened 2021-01-10
Audit celery task performance


rgossiaux:
We saw a big backlog in celery during the playtest. We should audit our tasks. Probably our workers are just idling waiting on API responses in which case we should be doing more tasks concurrently. Can also prioritize sheet creation over everything else for example.

rgossiaux:
We don't want to increase concurrency too much because we have pretty low connection limits from Heroku. And @rawxfish fixed the rate limiting that was slowing us down + prioritized sheet creation, so I think we should be ok at this point. I did a mild stress test earlier this evening and celery was pretty responsive with the rate limit change, and we don't really make that many tasks to begin with. So I'm going to close for now.

PR #374, opened 2021-01-11
remove update_meta_sheet_feeders rate limit

rawxfish:
Let's see what happens, somewhat related to #353


erwa (approved):
Looks like something worth trying. One task per minute for the update_meta_sheet_feeders seems too slow if we are updating a bunch of puzzles in a short period (e.g.: a new round of puzzles is released and we do a lot of feeder/meta organizing).

PR #375, opened 2021-01-11
Use sheet redirect endpoint on the frontend

rgossiaux:
Makes /puzzles responses almost 25% shorter. Hard to say precisely how
it affects the response time (too variable) but it appears to be a nice
improvement as well which is what I would expect. The only downside is
that opening a sheet is now a little slower since we have to do the
redirect.

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -100,6 +100,7 @@ class PuzzleSerializer(serializers.ModelSerializer):
     chat_room = ChatRoomSerializer(required=False)
     tags = PuzzleTagSerializer(required=False, many=True)
     guesses = serializers.SerializerMethodField()
+    sheet = serializers.SerializerMethodField()
     # Have to specify this explicitly for validate_url to run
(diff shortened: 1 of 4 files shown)


api/serializers.py
```diff
@@ -100,6 +100,7 @@ class PuzzleSerializer(serializers.ModelSerializer):
     chat_room = ChatRoomSerializer(required=False)
     tags = PuzzleTagSerializer(required=False, many=True)
     guesses = serializers.SerializerMethodField()
+    sheet = serializers.SerializerMethodField()
```
akirabaruah, line 103: nit: Should we call this "has_sheet" instead to clearly imply a boolean? I initially interpreted this as a sheet URL before seeing the test.

akirabaruah (approved):
Looks great!

api/serializers.py
rgossiaux: Yeah we should, good call

PR #377, opened 2021-01-11
Lower worker connection pool 2 => 1

rgossiaux:
Looks like this is one connection pool per worker process (and multiple
worker processes run at once) so keeping this small. Can also try None
for no connection pool (new connection every time) if we have to.


rgossiaux:
Forgot to lower the redis one but I will do that as well in this PR.

After merging this I will try stress testing again but if the connection
counts look too high I think we should just pay for the next highest redis
tier (which I recently turned on for debugging purposes) which is only a
couple dollars to run through the hunt. More important issues to worry
about than something than can be solved with a couple dollars.

rawxfish (approved):

PR #378, opened 2021-01-11
Add progress chart

bcai2:
image

* Progress chart (#200), built with [Chart.js](https://www.chartjs.org/), now appears on the stats page, using backend data gathering in `chart_utils.py`. Features include:
  * Meta solves are highlighted as blue circles that can be moused over for their names/solve times.
  * Times of ticks along the x-axis are alwas given in a format such as `Fr 4:00 pm`, in the local timezone. (Not sure if I should standardize this as ET or hunt time, but I don't think that would be better or necessary.) The minute of every tick time will be a multiple of 15.
  * Ticks are limited in count so as to not make the chart too messy.
  * Grid dimensions are based on the size of the page window.
  * Chart extends from the start time to either the current time or the end time (if specified), whichever is earlier.
  * Chart creation/display is disabled if the start time isn't given or if the current time is before the start time.

(Remaining todo: make the chart and stats page as a whole update every 10 seconds or so. This might not be super necessary and would be the last remaining major step, but it would be nice.)


hunts/chart_utils.py
```diff
@@ -0,0 +1,41 @@
+from django.utils import timezone
+
+from .models import Hunt
+from puzzles.models import Puzzle
+
+
+def can_use_chart(hunt):
+    if not hunt.start_time or timezone.now() <= hunt.start_time:
```
asdfryan, line 8: Can we use first puzzle creation time as a proxy when `hunt.start_time` doesn't exist?

hunts/chart_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_chart_data(hunt):
+    labels = ["Start"]
+    times = [hunt.start_time.isoformat()]
+    counts = [0]
+    is_meta = [False]
+
+    current_time = timezone.now()
+    if hunt.end_time:
+        current_time = min(current_time, hunt.end_time)
```
asdfryan, line 21: Nit: Naming this `current_time` is confusing to read.

This corresponds to the chart end time right? Can we rename to `chart_end_time`?

hunts/chart_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    for i, puzzle in enumerate(sorted_puzzles):
+        if puzzle.solved_time() > current_time:
+            solved_counts = i
+            break
+        labels.append(puzzle.name)
+        times.append(puzzle.solved_time().isoformat())
+        counts.append(i + 1)
+        is_meta.append(puzzle.is_meta)
+
+    labels.append("Now")
```
asdfryan, line 36: Nit: This should only be `Now` if `timezone.now() <= end_time` right, and `End` otherwise.

hunts/chart_utils.py
```diff
@@ -0,0 +1,41 @@
...
+from puzzles.models import Puzzle
+
+
+def can_use_chart(hunt):
+    if not hunt.start_time or timezone.now() <= hunt.start_time:
+        return False
+    return True
+
+
+def get_chart_data(hunt):
```
asdfryan, line 13: Can you add comments for what the output format is expected to be?

hunts/templates/stats.html
```diff
@@ -12,12 +13,117 @@ <h1 class="text-left">Hunt Statistics</h1>
...
+            time: {
+              unit: "minute",
+              stepSize: 15,
+              displayFormats: {
+                day: "dd",
+                hour: "dd hA",
+                minute: "dd h:mm a",
+              },
+              parser: function (timeInput) {
+                return moment(timeInput) //.utcOffset("-0500") // Eastern Time
```
asdfryan, line 83: Is this comment intentional?

bcai2: Somewhat. I was experimenting with displaying timezones for the actual chart. I'll most likely remove it for the final version.

hunts/templates/stats.html
```diff
@@ -12,12 +13,117 @@ <h1 class="text-left">Hunt Statistics</h1>
...
+              }
+            },
+            ticks: {
+              maxTicksLimit: 11,
+            },
+          }],
+          yAxes: [{
+            ticks: {
+              beginAtZero: true,
+              callback: function(value) {if (value % 1 === 0) {return value;}},
```
asdfryan, line 93: Why is this callback needed?

asdfryan (commented):
Very nice! Added a few nitty comments.

A few questions:
- Does this chart auto-refresh? It would be cool to have a cardinality stream that shows this chart over time (auto-refreshing every minute or so)
- What do you think of charting unlocks on top of solves? (with a different color)

hunts/templates/stats.html
bcai2: It's so we don't have solved value (y-axis) ticks that are decimals (e.g., 3.375).

asdfryan: Ah okay. Can you add a comment that this is what this is for? It's not intuitive from reading the code that this is the reason why.

hunts/chart_utils.py
```diff
@@ -0,0 +1,41 @@
+from django.utils import timezone
+
+from .models import Hunt
+from puzzles.models import Puzzle
+
+
```
asdfryan, line 6: Another nit: both functions in this file are unit-testable. Could you add tests?

asdfryan: Ah never mind I see that you already did. 👍

bcai2: Tests have been added!

hunts/views.py
```diff
@@ -57,6 +63,8 @@ def stats(request, hunt_slug):
         "minutes_per_solve": minutes_per_solve,
         "hunt_name": hunt.name,
         "hunt_slug": hunt.slug,
+        "use_chart": use_chart,
```
asdfryan, line 66: Is this necessary? Can't we just use existence of chart_data?

hunts/views.py
```diff
@@ -48,6 +49,11 @@ def stats(request, hunt_slug):
     solves_per_hour = hunt.get_solves_per_hour()
     minutes_per_solve = hunt.get_minutes_per_solve()
 
+    use_chart = can_use_chart(hunt)
+    chart_data = None
+    if use_chart:
```
asdfryan, line 54: Nit: I don't think you need `use_chart` -- can just call `if can_use_chart(hunt):` directly IMO

bcai2:
> Very nice! Added a few nitty comments.
> 
> A few questions:
> 
> * Does this chart auto-refresh? It would be cool to have a cardinality stream that shows this chart over time (auto-refreshing every minute or so)
> * What do you think of charting unlocks on top of solves? (with a different color)

- Not yet, but Chart.js has easily accessible ways to update a chart that's already on the page (with smooth animations, too). The implementation might be tricky and I'd probably be doing this with the stats page update, so I'm thinking of including that feature in a future issue/PR if I have enough time to implement it. (At this rate, hopefully!)
- Totally doable. Aesthetically, though, the unlocks might greatly outpace our solves, which might squish the solves line down greatly (and make it look pretty flat), so I'm not sure if it would look good.

I'm also thinking of recoloring the solve line green instead of making the color a reference, just to make it in line with our existing color scheme. (If we were to add the unlock line, that would be gray.)

asdfryan:
> > Very nice! Added a few nitty comments.
> > A few questions:
> > 
> > * Does this chart auto-refresh? It would be cool to have a cardinality stream that shows this chart over time (auto-refreshing every minute or so)
> > * What do you think of charting unlocks on top of solves? (with a different color)
> 
> * Not yet, but Chart.js has easily accessible ways to update a chart that's already on the page (with smooth animations, too). The implementation might be tricky and I'd probably be doing this with the stats page update, so I'm thinking of including that feature in a future issue/PR if I have enough time to implement it. (At this rate, hopefully!)
> * Totally doable. Aesthetically, though, the unlocks might greatly outpace our solves, which might squish the solves line down greatly (and make it look pretty flat), so I'm not sure if it would look good.
> 
> I'm also thinking of recoloring the solve line green instead of making the color a reference, just to make it in line with our existing color scheme. (If we were to add the unlock line, that would be gray.)

Awesome.

I think for the purposes of Mystery Hunt, the unlocks shouldn't significantly outpace solves.

asdfryan (approved):
Thanks! A few more nits.

hunts/chart_utils.py
```diff
@@ -0,0 +1,63 @@
...
+def can_use_chart(hunt):
+    if not hunt.start_time:
+        if hunt.get_num_unlocked() == 0:
```
asdfryan, line 9: This can be shortened as `return hunt.get_num_unlocked() > 0`

hunts/chart_utils.py
```diff
@@ -0,0 +1,63 @@
...
+            return False
+        else:
+            return True
+
+    if timezone.now() <= hunt.start_time:
```
asdfryan, line 14: `return timezone.now() > hunt.start_time`

bcai2:
The unlocked line has been successfully added!

image

All tooltips are disabled for the unlocks line, and a few color changes (to lines and fill areas) have been made in line with the existing color scheme.

hunts/chart_utils.py
```diff
@@ -0,0 +1,69 @@
...
+# times: Solve times of solved puzzles, in ISO 8601 format strings
+# counts: Total solves/unlocks at the time of that puzzle's solve/unlock (inclusive)
+# is_meta: Whether the solved puzzle is meta (list of booleans)
+def get_chart_data(hunt, unlocks=False):
+
+    # if start_time not given, use first puzzle creation time
+    # (use can_use_chart() to check beforehand)
+    chart_start_time = hunt.start_time
+    if not chart_start_time:
+        chart_start_time = hunt.puzzles.earliest("created_on").created_on
```
rgossiaux, line 25: Can't .earliest return None? Does this throw an error if there are no puzzles in the hunt? I guess the idea is you should call `can_use_chart` first before calling this function but maybe there should be some handling at the start of this function too

bcai2: Yeah, the idea was that `can_use_chart` would always be called before `get_chart_data`, but some handling within the function would also be an option. If we were to do that, I'd probably get rid of `can_use_chart` altogether and reincorporate it into the body of `get_chart_data`.
(Actually, that does seem more sensible, now that I think about it.)

PR #381, opened 2021-01-12
Disable celery heartbeat, gossip, mingle

rgossiaux:
Some suggestions online to disable these for keeping redis connection
counts down as they all add extra connections. Gossip is only used for
clock sync, mingle is only for some inter-worker sycnhronization
(especially for letting new workers know about unprocessed revoked
tasks--not something we need to worry about as our worker pool will be
constant and we won't be revoking tasks), heartbeat is useless as we
only have one redis node.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 release: python manage.py migrate --noinput
 web: gunicorn smallboard.wsgi --log-file -
-worker: celery -A smallboard worker -l INFO
+worker: celery -A smallboard worker -l INFO --without-heartbeat --without-gossip --without-mingle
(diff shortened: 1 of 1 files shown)


rawxfish (approved):
Looks good, did the friendly strangers online say what we should put for `--concurrency`?

rgossiaux:
> Looks good, did the friendly strangers online say what we should put for `--concurrency`?

Lol from the perspective of minimizing redis connections it should be as small as possible but I'm hoping we don't have to mess with that, don't want everything to get backed up because of one slow task.

PR #382, opened 2021-01-12
Lock AUTOGENERATED tab (#366)

rawxfish:
- lock AUTOGENERATED tab
- instead of deleting existing AUTOGENERATED tab and creating new one, clear existing tab and repopulate. The old behavior would kick people off the tab which was a little annoying
- attempt to refactor code.


google_api_lib/task.py
```diff
@@ -126,155 +165,156 @@ def update_meta_sheet_feeders(self, puzzle_id):
...
+            {
+                "addProtectedRange": {
+                    "protectedRange": {
+                        "range": {
+                            "sheetId": sheet_id,
+                        },
+                        "description": "Autogenerated page",
+                        "warningOnly": True,
+                    }
+                }
```
rgossiaux, line 312: Does this block smallboard from editing the sheet? I guess maybe not because it's warningOnly? I think you can maybe put the requesting user in `editors` and then set this to be False for extra strictness.

rawxfish: Extra strictness added!

rgossiaux (approved):
Nice, thanks! lgtm

Issue #373, opened 2021-01-11
Rename sheet when puzzle is deleted


rawxfish:
So people realize if the sheet is abandoned.

PR #383, opened 2021-01-12
Rename sheet when puzzle is deleted (#373)

rawxfish:


rgossiaux (commented):
What do you think about combining this with `update_meta_sheets_pre_delete` in puzzles/signals/handlers.py (and maybe renaming that function)? I've been trying to keep some of this logic around updating sheets in those signal handlers to some extent although not been super strict about it.

rawxfish:
oh yeah we should do that instead

rgossiaux (approved):
Ty! nit: maybe rename `update_meta_sheets_pre_delete` to just `update_sheets_pre_delete`

PR #384, opened 2021-01-12
prioritize sheet creation (#353)

rawxfish:


google_api_lib/utils.py
```diff
@@ -7,6 +8,12 @@
 from celery import Task
 
 
+class Priority(Enum):
```
rgossiaux, line 11: Should we move this somewhere else? Admittedly not really sure where.

rgossiaux (commented):
Do we need to add `queue_order_strategy`? https://docs.celeryproject.org/en/stable/userguide/routing.html#redis-message-priorities

google_api_lib/utils.py
rawxfish: moved to settings.py

rawxfish:
> Do we need to add `queue_order_strategy`? https://docs.celeryproject.org/en/stable/userguide/routing.html#redis-message-priorities

cry

smallboard/settings.py
```diff
@@ -267,13 +267,23 @@
 
 
 # Celery settings
+from enum import Enum
+
+
+class Priority(Enum):
```
rgossiaux, line 273: Nit: maybe something like `TaskPriority`

rgossiaux (approved):
Ty!

PR #385, opened 2021-01-12
Fix meta tag updates

pcvera:


rgossiaux (commented):
Nice, thank you!!!

I think fetching all the puzzles can get a little slow (~500ms and maybe it'll slow more under load) so I wonder if this is going to add some visual lag (though that's still better than the confusing intermediate state that appears now). What do you think of having the addPuzzleTag / deletePuzzleTag calls return the rows that need to update so we don't need to make a full refetch? I guess the only things that need to update are the current puzzle + the puzzle corresponding to the meta tag if the tag is a meta.

pcvera:
I started looking at it after the call, I won't have time to do it tonight but I'll take a longer look tomorrow morning. It looks like in the case of adding/removing a meta tag I'll have to fetch that puzzle record too, I saw that there's already a call to `Puzzle.objects.get(name=tag.name, hunt=puzzle.hunt)`, but that doesn't return with the same type of puzzle object that the `PuzzleSerializer` is expecting. If you'd like to give me a hint, I think the best way to approach fetching the meta is what I'm most hurting for.

rgossiaux:
> I started looking at it after the call, I won't have time to do it tonight but I'll take a longer look tomorrow morning. It looks like in the case of adding/removing a meta tag I'll have to fetch that puzzle record too, I saw that there's already a call to `Puzzle.objects.get(name=tag.name, hunt=puzzle.hunt)`, but that doesn't return with the same type of puzzle object that the `PuzzleSerializer` is expecting. If you'd like to give me a hint, I think the best way to approach fetching the meta is what I'm most hurting for.

I think Puzzle.objects.get should return a Puzzle object and PuzzleSerializer should be able to handle that-- I think if you do something like

```
puzzle = none
meta = None
with transaction.atomic():
    # ....
    # meta = Puzzle.objects.get(name = tag.name, hunt=puzzle.hunt)
    # ....

if meta:
    return Response([PuzzleSerializer(puzzle).data, PuzzleSerializer(meta).data])
else:
    return Response([PuzzleSerializer(puzzle).data])
```

that should probably be enough for the `destroy` call (+ a similar change to the `create` call). Then on the front-end the thunks can return this response instead of the response from `fetchPuzzles`, and you already updated the reducers to expect a list of puzzles so nothing to change there. Lmk if that doesn't work or make sense.

api/views.py
```diff
@@ -289,11 +290,16 @@ def destroy(self, request, pk=None, **kwargs):
                 transaction.on_commit(
                     lambda: chat.tasks.handle_tag_removed.delay(puzzle.id, tag.name)
                 )
-
-        return Response(PuzzleSerializer(puzzle).data)
+        if meta == None:
```
rgossiaux, line 293: in python you should usually check `is None` instead of `== None`; in this case just `if meta` / `if not meta` is the most "pythonic" imo

api/views.py
```diff
@@ -340,4 +346,9 @@ def create(self, request, **kwargs):
                     lambda: chat.tasks.handle_tag_added.delay(puzzle.id, tag.name)
                 )
 
-        return Response(PuzzleSerializer(puzzle).data)
+        if meta == None:
```
rgossiaux, line 349: ditto

rgossiaux (approved):
Yay, thank you!!

rgossiaux:
I rudely just changed it myself because we want to merge everything now.

Thanks for fixing this!!!

PR #387, opened 2021-01-12
Discord mention roles when their tag was created

rgossiaux:
Feels a little janky right now but hey it works. Not sure the best way of structuring it. Originally I wanted to make this discord-specific but the discord_lib file isn't a django app / can't depend on django model stuff because it's imported in settings before models are available.

There's no way in the UI to edit these roles right now, I'm relying on the Django admin page.

Should probably also add a special announcement for @meta


api/views.py
```diff
@@ -281,6 +281,13 @@ def destroy(self, request, pk=None, **kwargs):
             if not tag.puzzles.exists():
                 tag.delete()
 
+            if puzzle.chat_room:
+                transaction.on_commit(
+                    lambda: puzzle.chat_room.get_service().handle_tag_removed(
```
asdfryan, line 286: Hmm do you think `handle_tag_removed` makes more sense to be defined at the ChatRoom level?

I think initially, Akira and I agreed to not have the service exposed in the views (I know we kind of broke that rule with `announce()` but I kinda want to make that a static or class method under ChatRoom).

Let me know what you think?

asdfryan (commented):
Nice! Thanks for implementing this.

I had one small comment.

api/views.py
rgossiaux: Sure, happy to move it to ChatRoom to make it more consistent with the existing code. I think the structure right now feels a little unintuitive to me because this isn't really specific to the particular chat room that's associated with the puzzle but rather with the Discord server as a whole, but we can worry about that kind of thing later.

asdfryan (approved):

Issue #379, opened 2021-01-12
[UI] Stats UI improvements and additions


bcai2:
From playtest feedback.

- [x] Link to stats page should be closer to the other stats instead of in the corner
- [x] List of meta solves on stats page
- [x] Recent solve rate stats (e.g. past 6 hours)
- [x] Progression (# solves + # puzzles with solved metas)
- [x] Metas unsolved stat (optional, IMO, but I'll see if I can make it fit nicely)

Issue #322, opened 2021-01-09
Autotag puzzles as lo-pri if their metas get solved


asdfryan:

rgossiaux:
I like this idea a lot. We'll just need to be a little careful on the implementation details since puzzles can have multiple metas & can have several levels of nesting (if my metameta gets solved i should be low pri too even if my meta is still unsolved), and I guess we'll want to un-tag things if meta relationships change (maybe we discover a new meta that this puzzle feeds into) or the meta becomes unsolved (eg someone entered the answer for the wrong puzzle and corrects it).

asdfryan:
Hmm you're right that there are a bunch of edge cases to consider with multiple meta support.

More generically, I think the following invariant must be true: puzzle should be auto-tagged as lo-pri if and only if ALL of its metas have been solved, OR if all of its metametas have been solved, etc.

Any operation that makes that statement newly true should auto-tag. Inversely, any operation that makes that statement newly false should untag.

This could make implementation a bit tricky -- especially when adding new meta relationships or "unsolving" a meta, you'd have to check all the way down the feeder tree to see if things should be untagged.

Alternatively, we could limit the scope of this feature to just a single level (i.e. just check the meta and ignore metametas) and it'd make the implementation way simpler.

Let me know if you disagree.

erwa:
I think it's reasonable to just check a single level (so mark as lo-pri if all of a puzzle's direct metas have been solved). I think the check to decide whether to autotag could be done whenever a meta is marked as solved: For each feeder, check if all of its direct metas are solved, and if so, autotag the feeder as lo-pri.

But if someone then manually deletes the lo-pri tag, I don't think it should automatically be reapplied because we may suspect a puzzle might feed into another meta but we haven't unlocked that meta or aren't sure which meta it is.

rgossiaux:
I think for modeling multiple levels of the relationship it's actually a
straightforward condition: a puzzle should be low pri if all its parents
are either solved or low pri. I'm not sure how that's best computed but it
translates directly to a recursive CTE in SQL I think.

Supporting manually overriding the priority is probably a good idea too.

PR #389, opened 2021-01-12
Stats UI improvements from feedback (#379)

bcai2:
Implemented suggested additions and feedback from the playtest. (#379)

* Link to stats page moved closer to the other stats instead of being in the corner
* List of meta solves on stats page
* Recent solve rate stats (e.g. past 6 hours)
* Progression (# solves + # puzzles with solved metas)
* Metas unsolved stat
* Stats page "lazily" refreshes every 60 seconds (via meta refresh)
* Added unit tests for new stats computed (recent solve rates, metas unsolved, data for meta solve chart, progression)

image

image


bcai2:
Note: the times in the Metas Solved table are in the default timezone (ET), regardless of which local timezone the user's in. I know this is inconsistent with the desired chart behavior (displayed in local time), but Django makes this *really* annoying. 

I've added an ET descriptor label. This seems low-priority, so I'll table this for now and get back to it once I integrate the chart and implement page refreshing/updating. Hopefully I'll be able to amend this before too long.

image

hunts/models.py
```diff
@@ -47,27 +48,72 @@ def get_num_unlocked(self):
...
+                        solved_puzzle_ids.add(feeder.id)
+                        puzzles_to_check.append(feeder)
+
+        return len(solved_puzzle_ids)
+
+    # Returns a list of solved meta names and solve times in [name, time] pairs.
+    # Solve times are given in dd hh:mm (am/pm) format (e.g., Fr 4:00 pm).
+    def get_meta_solve_list(self):
+        solved_metas = self.puzzles.filter(Q(status=Puzzle.SOLVED), Q(is_meta=True))
+        sorted_solved_metas = sorted(solved_metas, key=lambda x: x.solved_time())
```
rgossiaux, line 74: I think this could also make a bunch of db queries, probably fine for now but maybe add a TODO for it (I think we are going to want to prefetch before calling solved_time() on each meta since each call makes a new lookup)

hunts/models.py
```diff
@@ -47,27 +48,72 @@ def get_num_unlocked(self):
...
+        interval_start = self.start_time
+        if recent:
+            interval_start = max(interval_start, interval_end - timedelta(hours=6))
 
         solved = self.get_num_solved()
-        hours_elapsed = (current_time - self.start_time).total_seconds() / 3600
+        if recent:
+            solved_puzzles = self.puzzles.filter(Q(status=Puzzle.SOLVED))
+            solved = len(
+                [x for x in solved_puzzles if x.solved_time() >= interval_start]
```
rgossiaux, line 95: This could have similar perf issues, let's put another TODO

rgossiaux (changes requested):
Nice! Comments inline but mostly lgtm

hunts/models.py
```diff
@@ -47,27 +48,72 @@ def get_num_unlocked(self):
...
+        puzzles_to_check = [p for p in solved_puzzles]
+
+        while len(puzzles_to_check) > 0:
+            next_puzzle = puzzles_to_check.pop()
+            if next_puzzle.is_meta:
+                for feeder in next_puzzle.feeders.all():
+                    if feeder.id not in solved_puzzle_ids:
+                        solved_puzzle_ids.add(feeder.id)
+                        puzzles_to_check.append(feeder)
+
```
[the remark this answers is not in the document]
bcai2: Would this get all puzzles below a solved puzzle in the "puzzle tree" as well? (Including a feeder of that puzzle's feeder, e.g.)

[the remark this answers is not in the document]
bcai2: Upon further thought, I think I'm just going to go with your suggestion. If there is a difference, I can't see it mattering until very late in the hunt (when we've solved the final metameta/puzzle in the puzzlehunt or something), so I'm just going to go with "is solved or has a meta that's solved".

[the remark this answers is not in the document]
rgossiaux: Ah I see, I missed that you were traversing all the way down the tree. 

I think this is similar to #322 in that I think ideally both want the same set of puzzles (every "unnecessary" puzzle, where a puzzle is unnecessary if it is solved or if all its metas are unnecessary). 

I think the right way to do this in the DB is with a recursive CTE, something like (pseudocode):

```
WITH RECURSIVE unnecessary_puzzles (id, metas) AS (SELECT id, metas FROM puzzles WHERE puzzle is solved UNION ALL SELECT id, metas FROM puzzles WHERE metas in unnecessary_puzzles) SELECT * FROM unnecessary_puzzles```

though idk if we want to go down that path or not. Anyway what you did is fine, I think you can alternatively prefetch `feeders` in your initial query and it should prefetch everything into one query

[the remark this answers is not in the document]
Comment (bcai2) [reply hunts/models.py:67]: I got it to work with the following recursive call:
```
WITH RECURSIVE progression_puzzles (id) AS (
  SELECT id
  FROM puzzles_puzzle
  WHERE status = 'SOLVED' AND hunt_id = %d
  UNION
  (
    -- workaround psql error: "recursive reference in a subquery"
    WITH progression_ids AS (
        SELECT id FROM progression_puzzles
    )
    SELECT Metas.from_puzzle_id
    FROM puzzles_puzzle_metas Metas
    WHERE Metas.to_puzzle_id IN (SELECT * FROM progression_ids)
  )
)
SELECT id FROM progression_puzzles
```
This might be the hackiest thing I've done all week.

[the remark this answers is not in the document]
Comment (bcai2) [reply hunts/models.py]: Update: This doesn't work with SQLite (which Github's currently using to test), so I used an `INNER JOIN` instead.
```
WITH RECURSIVE progression_puzzles (id) AS (
            SELECT id
            FROM puzzles_puzzle
            WHERE (status = 'SOLVED' AND hunt_id = %s)
            UNION
            SELECT Metas.from_puzzle_id
            FROM puzzles_puzzle_metas Metas
            INNER JOIN progression_puzzles ON progression_puzzles.id = Metas.to_puzzle_id
        )
        SELECT id FROM progression_puzzles
```

**Comment by bcai2** [comment]
THANK GOD

File: hunts/models.py (hunk 1 above)
Comment (bcai2) [reply hunts/models.py]: Turns out that I can use `.annotate(custom_condition_here)` in the Django ORM to replicate `solved_time()` without making too many DB queries from calling `solved_time()` over and over again! I've replaced every instance of calling the latter too many times with the former, which should hopefully amend any future problems.

File: hunts/models.py [hunk 4]
```diff
@@ -47,27 +48,80 @@ def get_num_unlocked(self):
     def get_num_metas_solved(self):
         return self.puzzles.filter(Q(status=Puzzle.SOLVED), Q(is_meta=True)).count()
 
-    def get_solves_per_hour(self):
-        current_time = timezone.now()
+    def get_num_metas_unsolved(self):
+        return self.puzzles.filter(~Q(status=Puzzle.SOLVED), Q(is_meta=True)).count()
+
+    # Gets the number of puzzles that are either solved or feed into a solved meta.
```
rgossiaux, line 54: nit: maybe "only feed into solved metas" for clarity

hunts/models.py
```diff
@@ -47,27 +48,80 @@ def get_num_unlocked(self):
...
+        WITH RECURSIVE progression_puzzles (id) AS (
+            SELECT id
+            FROM puzzles_puzzle
+            WHERE (status = 'SOLVED' AND hunt_id = %s)
+            UNION
+            SELECT Metas.from_puzzle_id
+            FROM puzzles_puzzle_metas Metas
+            INNER JOIN progression_puzzles ON progression_puzzles.id = Metas.to_puzzle_id
+        )
+        SELECT id FROM progression_puzzles
```
asdfryan, line 66: Does SELECT COUNT(id) ... not work? We only care about the number of IDs right? This is also more efficient.

hunts/models.py
```diff
@@ -47,27 +48,80 @@ def get_num_unlocked(self):
...
+        """
+        progression_ids = Hunt.objects.raw(query, [str(self.pk)])
+        return len(list(progression_ids))
```
rgossiaux, line 69: Can we have the query select the whole puzzle row and then return the RawQuerySet of Puzzle objects? So SELECT * and then returning the Hunt.objects.raw result. Then the stats caller can call len(list(result)). I think this way we can reuse this function (which is computing a pretty interesting set of puzzles) for other things

rgossiaux: Lol I see @asdfryan and I both suggested different things. I wanted to reuse this for #322 or something along those lines

hunts/models.py
rgossiaux: Nice, thank you!! .annotate is definitely how you are supposed to do this in Django, good find. I think this may be slightly different (looks like it looks at the max time of all guesses instead of all correct guesses) but that's not a big deal (though maybe we should add a comment or TODO, in case we ever go back to an answer queue world)

bcai2: Looks like there are `Case` and `When` functions that allow for conditionally choosing. I'll look into those to give us the functionality that we want, hopefully.

hunts/models.py
bcai2: I'll implement this and use len(list(result)) when passing into the page.

rgossiaux (approved):
Thank you!!!

Issue #349, opened 2021-01-10
Better support for google drawing, jamboard, etc


rgossiaux:
We should link a Google Drive folder so that people can create drawings/jamboards/etc and have the permissions work out. Maybe a subfolder of the main folder for less clutter.

PR #393, opened 2021-01-13
per hunt google permissions and drives

rawxfish:
I don't know what came over me to think it would be a good idea to do this as apart of #349. It is not. This does not work. Will revisit after this weekend.


google_api_lib/utils.py
```diff
@@ -5,12 +5,44 @@
...
+    return m.group(1)
+
+
+def extract_id_from_drive_folder_url(url):
+    """
+    Assumes `url` is of the form
+    https://drive.google.com/drive/u/2/folders/<ID>/..
+    and returns the <ID> portion
+    """
+    pattern = ".*drive\.google\.com\/drive\/u\/2\/folders\/([-\w]+).*"
```
erwa, line 38: I believe these `/u/<n>/` URLs mean you are logged into multiple Google accounts -- /u/1 is the first account you're logged in to, /u/2 is the second, etc. This is going to be different per person depending on what accounts they're logged in to.

I think you want to use the canonical drive id, which should look something like https://drive.google.com/drive/folders/<id>...

(You can navigate to a folder in Drive and Get Link to get the canonical link)

erwa (commented):
Looks like a good start in supporting per-hunt Google Drive folders and permissions! Schema changes look good. After this, the next step would be to add APIs and update the frontend to support configuring the Google info when creating a new hunt.

One comment about extracting the id from the Google Drive URL.

hunts/views.py
```diff
@@ -21,11 +22,30 @@ def index(request):
...
+                    form.cleaned_data["google_drive_hunt_folder_url"]
+                )
+                sheets_template_file_id = utils.extract_id_from_sheets_url(
+                    form.cleaned_data["google_sheets_template_file_url"]
+                )
+                google_ids = GoogleInfo(
+                    hunt_folder_id,
+                    autogenerated_sheets_folder_id,
+                    human_files_folder_id=None,
+                    sheets_template_file_id=None,
```
rgossiaux, line 40: I think you're passing the wrong ones here.

I wonder if actually we could get Smallboard to manage this stuff: upon creating a new hunt, we automatically create the folders for it & copy in our template file

hunts/models.py
```diff
@@ -1,18 +1,52 @@
...
+        super(GoogleInfo, self).save(*args, **kwargs)
+
 
 class Hunt(models.Model):
     name = models.CharField(max_length=128)
     url = models.URLField()
     created_on = models.DateTimeField(auto_now_add=True)
     start_time = models.DateTimeField(default=None, blank=True, null=True)
     end_time = models.DateTimeField(default=None, blank=True, null=True)
+    google_ids = models.OneToOneField(
```
rgossiaux, line 44: Might rename this to `google_config` or something, I thought `ids` meant users at first

rgossiaux (commented):
thanks for working on this!

rawxfish:
#538 does all of this, so closing

PR #394, opened 2021-01-13
add link to google drive folder (#349)

rawxfish:
- Look for a folder in the folder given by env variable called "FOLDER FOR HUMANS". Create it if it does not exist
- Add an url redirect to human folder. This is hunt specific since we'll eventually want it that way anyways.
- add button on hunt page:

image


erwa (commented):
Could the human drive folder just be a folder under the `GOOGLE_DRIVE_HUNT_FOLDER_ID` that's automatically created by Small Board? Just trying to avoid adding another config if not necessary. This way, when setting up a new hunt instance, you just need to configure a single Google Drive folder, and if permissions are set properly on that, they should propagate to subfolders and files. Adding another folder id increases the likelihood we forget to configure it and of having misconfigured permissions. Not a huge issue to have two drive ids but this just seems like something that could be consolidated.

erwa (commented):
Ignore my previous comment; I didn't look at your code closely enough. It seems you are just specifying the name of the human folder and creating it under the Drive folder.

smallboard/settings.py
```diff
@@ -233,16 +233,21 @@
...
     )
     logger.info(
         "Whitelisted emails: " + str(SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_EMAILS)
     )
+    GOOGLE_HUMAN_DRIVE_HUNT_FOLDER_URL = ""
+    GOOGLE_HUMAN_DRIVE_HUNT_FOLDER_URL = sync_task.get_human_drive_folder(
+        GOOGLE_DRIVE_HUNT_FOLDER_ID
+    )
 else:
+    GOOGLE_HUMAN_DRIVE_HUNT_FOLDER_URL = ""
```
erwa, line 250: Instead of repeating `GOOGLE_HUMAN_DRIVE_HUNT_FOLDER_URL = ""`, can just initialize before the `if GOOGLE_API_AUTHN_INFO is not None` and remove it from the if and else blocks.

erwa (approved):
Looks good to me. Minor style comment.

hunts/src/huntSlice.js
```diff
@@ -17,6 +17,7 @@ export const huntSlice = createSlice({
     [fetchHunt.fulfilled]: (state, action) => {
       state.id = action.payload.id;
       state.name = action.payload.name;
+      state.has_drive = action.payload.has_drive;
```
rgossiaux, line 20: Should add this to initialState I think.

package.json
```diff
@@ -35,6 +35,7 @@
     "react-dom": "^17.0.1",
     "react-redux": "^7.2.2",
     "react-table": "^7.6.2",
+    "react-tooltip": "^4.2.11",
```
rgossiaux, line 38: This works too but I think I have some small preference for using the tooltips in react-boostrap since we're using that for our other UI components: https://react-bootstrap.github.io/components/overlays/#tooltips

rgossiaux (commented):
Thank you for adding this!!

google_api_lib/sync_task.py
```diff
@@ -0,0 +1,61 @@
+from .utils import GoogleApiClientTask
+from celery import shared_task
+import logging
+
```
asdfryan, line 4: Rename to sync_tasks.py for convention?

rawxfish:
now it look like this

image

yarn.lock
```diff
@@ -2640,6 +2640,14 @@ react-table@^7.6.2:
   resolved "https://registry.yarnpkg.com/react-table/-/react-table-7.6.2.tgz#b60932fa6d457c2bca0da49815cd6a8fe9451f77"
   integrity sha512-urwNZTieb+xg/+BITUIrqdH5jZfJlw7rKVAAq25iXpBPwbQojLCEKJuGycLbVwn8fzU+Ovly3y8HHNaLNrPCvQ==
 
+react-tooltip@^4.2.11:
```
rgossiaux, line 2643: Remove this? Looks like you're not touching package.json any more so you can undo your changes here too

PR #395, opened 2021-01-13
Prevent stale fetches from overwriting changes

### LINKED ISSUE #328: Prevent concurrent API call stale data bugs
Sometimes you can add a puzzle and then have it immediately disappear, or delete it and have it immediately reappear, etc. when the following sequence happens:

1) We begin fetching an update for all the puzzles
2) You make an API call and it succeeds.
3) The fetch finishes and returns.

Then the change you made in step 2 is overridden with the stale data returned in 3.

Two ways of making this better:

* The faster we can make fetching puzzles, the less often it'll happen.
* We can cancel/ignore stale requests if we make another one in the interim. The pending/fulfilled/rejected redux actions should make this manageable.

rgossiaux:
Fixes #328.

Thought about trying to retry the fetch but since we fetch every several
seconds anyway it didn't seem necessary (plus it could bunch up several
at once which could be undesirable).

### DIFF
--- a/hunts/src/puzzlesSlice.js
+++ b/hunts/src/puzzlesSlice.js
@@ -25,9 +25,10 @@ export const deletePuzzle = createAsyncThunk(
 
 export const fetchPuzzles = createAsyncThunk(
   "puzzles/fetchPuzzles",
-  async (huntId) => {
+  async (huntId, { getState }) => {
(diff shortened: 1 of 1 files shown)


erwa (approved):
Nice, a timestamp seems like a good way of always using the latest fetch

PR #396, opened 2021-01-13
Chat operations as Celery tasks

rgossiaux:
Figured I might as well do this now. The discord API is fairly
responsive but I'd noticed it was still introducing some lag to a lot of
interactions.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -18,6 +18,7 @@ from .serializers import (
     PuzzleTagSerializer,
 )
 import google_api_lib.task
+import chat.tasks
 
 import logging
 
@@ -67,13 +68,11 @@ class AnswerViewSet(viewsets.ModelViewSet):
                 answer.status = Answer.CORRECT
                 puzzle.answer = answer.text
                 if puzzle.chat_room:
-                    try:
-                        puzzle.chat_room.archive_channels()
-                        msg = f"{puzzle.name} has been solved with {answer.text}!"
-                        puzzle.chat_room.get_service().announce(msg)
-                        puzzle.chat_room.send_message(msg)
-                    except Exception as e:
-                        logger.warn(f"Chat operations failed with error: {e}")
+                    transaction.on_commit(
+                        lambda: chat.tasks.handle_puzzle_solved.delay(
+                            puzzle.id, answer.text
+                        )
+                    )
                 answer.save()
                 transaction.on_commit(
                     lambda: AnswerViewSet._maybe_update_meta_sheets_for_feeder(puzzle)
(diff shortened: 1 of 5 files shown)


chat/models.py
```diff
@@ -116,11 +116,11 @@ def send_message(self, msg):
             service = self.get_service()
             service.send_message(self.text_channel_id, msg)
 
-    def handle_tag_added(self, puzzle, tag):
-        self.get_service().handle_tag_added(self, puzzle, tag)
+    def handle_tag_added(self, puzzle, tag_name):
+        self.get_service().handle_tag_added(puzzle, tag_name)
```
rawxfish, line 120: What's the reason for having handle_tag_added jump around from tasks to the chat model back to discord_chat_service.py?

asdfryan: IMO our initial plan was not to expose the service on any views (or tasks). This is WAI IMO.

chat/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        puzzle.chat_room.unarchive_channels()
+    except Exception as e:
+        logger.warn(f"Chat operations failed with error: {e}")
+
+
+@shared_task
+def handle_tag_added(puzzle_id, tag_name):
+    puzzle = Puzzle.objects.get(id=puzzle_id)
+    puzzle.chat_room.handle_tag_added(puzzle, tag_name)
```
asdfryan, line 48: Should we also wrap this with try except? Same for removed.

asdfryan (approved):
Nice! Approving with 1 (somewhat unrelated) nit. I'd like to get this in ASAP so I can iterate on my discord issues on top of it.

chat/models.py
rawxfish: Gotcha, makes sense

asdfryan (changes requested):
Looks like some tests are failing.

api/views.py
```diff
@@ -224,16 +222,11 @@ def create(self, request, **kwargs):
...
-                    chat_room.get_service().announce(msg)
-                    chat_room.send_message(msg)
-                except Exception as e:
-                    logger.warn(f"Chat operations failed with error: {e}")
+                transaction.on_commit(
+                    lambda: chat.tasks.create_chat_for_puzzle.delay(puzzle.id)
+                )
             else:
                 logger.warn("Chat room not created for puzzle %s" % name)
-                chat_room = None
```
asdfryan, line 236: I think you have to keep this line. Otherwise line 231 will fail.

rgossiaux:
I'm working for the next 4-5 hours--happy to merge after that, but if you
need this in sooner to unblock would you mind fixing your nit and the test
(sorry about that) and merging it? I agree with your suggestions.

asdfryan (approved):
Nits/tests are now fixed. Merging this. :)

rgossiaux:
Thank you!!

PR #401, opened 2021-01-14
Rename channels when puzzles are renamed

asdfryan:


chat/tasks.py
```diff
@@ -58,3 +58,12 @@ def handle_tag_removed(puzzle_id, tag_name):
         puzzle.chat_room.handle_tag_removed(puzzle, tag_name)
     except Exception as e:
         logger.warn(f"Chat operations failed with error: {e}")
+
+
+@shared_task
+def handle_channel_rename(puzzle_id, new_name):
```
rgossiaux, line 64: nit: maybe `handle_puzzle_rename`

api/views.py
```diff
@@ -186,6 +186,15 @@ def partial_update(self, request, pk=None, **kwargs):
                     puzzle.status = data["status"]
                     puzzle.save()
 
+                if puzzle.chat_room and "name" in data:
+                    puzzle.chat_room.name = data["name"]
+                    puzzle.chat_room.save()
+                    transaction.on_commit(
+                        lambda: chat.tasks.handle_channel_rename.delay(
+                            puzzle.id, data["name"]
+                        )
```
rgossiaux, line 195: Not 100% sure about what the chat model expects but should we rename puzzle.chat_room after we actually change the discord channel name successfully in the task?

rgossiaux (commented):
Nice!

api/views.py
asdfryan: As discussed offline -- `chat_room.name` is actually duplicative with puzzle.name. We should probably figure out if we actually need this (and delete it if not).

asdfryan:
Addressed comments. Also am now checking that the new name and old name are different before attempting to rename the channel.

rgossiaux (approved):

PR #403, opened 2021-01-14
Send message upon sheet creation. Also rename task.py -> tasks.py

asdfryan:


google_api_lib/tasks.py
```diff
@@ -50,8 +51,12 @@ def create_google_sheets(self, puzzle_id, name, puzzle_url=None):
     sheet_url = response["webViewLink"]
     if puzzle_url:
         add_puzzle_link_to_sheet(puzzle_url, sheet_url)
-    puzzle = Puzzle.objects.filter(pk=puzzle_id).update(sheet=sheet_url)
+    puzzle = Puzzle.objects.get(pk=puzzle_id)
+    puzzle.sheet = sheet_url
+    puzzle.save()
```
rgossiaux, line 56: nit: i think the other way still works and is one fewer db query (update vs get + update, not 100% sure though)

rgossiaux (approved):
Nice, thanks!

google_api_lib/tasks.py
asdfryan: I needed the puzzle object to call `chat_room` to confirm existence.

rgossiaux: Oops yup, didn't read closely, sounds good

PR #404, opened 2021-01-14
Add clicking on tags

rgossiaux:
@dan-simon implemented this feature in #343; I just moved some logic into redux. Making a clean PR since I was a reviewer on the last one (not sure how to best do this in github) but see that one for the history.


erwa (approved):
Nice! Seems like a useful feature

PR #405, opened 2021-01-14
Announce answer changes

asdfryan:


asdfryan:
Screen Shot 2021-01-14 at 1 11 42 AM

chat/tasks.py
```diff
@@ -60,6 +60,16 @@ def handle_tag_removed(puzzle_id, tag_name):
         logger.warn(f"Chat operations failed with error: {e}")
 
 
+@shared_task
+def handle_answer_change(puzzle_id, old_answer, new_answer):
+    puzzle = Puzzle.objects.get(id=puzzle_id)
+    try:
+        msg = f"{puzzle.name}'s answer changed from {old_answer} to {new_answer}."
+        puzzle.chat_room.send_and_announce_message(msg)
```
erwa, line 68: Do we need to check that old_answer != new_answer before sending an announcement, or is that check already done somewhere else?

erwa (approved):
LGTM. One question

chat/tasks.py
asdfryan: It is not, AFAIK.

PR #408, opened 2021-01-14
Make status editable even after marked as SOLVED

asdfryan:
`SOLVED` is only listed as an option for solved puzzles. (i.e. it is only possible to "SOLVE" a puzzle by adding an answer)

Screen Shot 2021-01-14 at 12 26 33 PM
Screen Shot 2021-01-14 at 12 26 43 PM

### DIFF
--- a/hunts/src/StatusCell.js
+++ b/hunts/src/StatusCell.js
@@ -27,7 +27,7 @@ export default function StatusCell({ row, value }) {
           );
         }}
       >
-        {statuses_to_display.map(status_to_display => (
+        {statuses_to_display.map((status_to_display) => (
           <Dropdown.Item eventKey={status_to_display} key={status_to_display}>
             {status_to_display}
           </Dropdown.Item>
--- a/hunts/src/StatusCell.js
+++ b/hunts/src/StatusCell.js
@@ -7,8 +7,12 @@ import { updatePuzzle } from "./puzzlesSlice";
 export default function StatusCell({ row, value }) {
   const { id: huntId } = useSelector((state) => state.hunt);
   const dispatch = useDispatch();
-  if (["SOLVING", "STUCK", "EXTRACTION"].includes(value)) {
-    return (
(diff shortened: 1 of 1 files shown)


hunts/src/StatusCell.js
```diff
@@ -7,8 +7,12 @@ import { updatePuzzle } from "./puzzlesSlice";
 export default function StatusCell({ row, value }) {
   const { id: huntId } = useSelector((state) => state.hunt);
   const dispatch = useDispatch();
-  if (["SOLVING", "STUCK", "EXTRACTION"].includes(value)) {
-    return (
+  var statuses_to_display = ["SOLVING", "STUCK", "EXTRACTION"];
```
erwa, line 10: It seems the convention is to use `const` for arrays, even if the _contents_ of them are modified, as long as the array variable itself is not reassigned.

https://stackoverflow.com/questions/46553062/what-is-the-preferred-declaration-convention-for-objects-or-arrays-const-or-let

hunts/src/StatusCell.js
```diff
@@ -7,8 +7,12 @@ import { updatePuzzle } from "./puzzlesSlice";
...
+  if (value == "SOLVED") {
```
erwa, line 11: Is `value` always a string here? If so, we can use `===`

erwa (approved):
Nice! Just some minor nits.

hunts/src/StatusCell.js
asdfryan: Thanks -- didn't know that. :)

rgossiaux (commented):
On my phone but is there a way to mark it back as solved once if you un-mark it? Maybe we should add SOLVED to the options if row.original.answers or whatever it's called has at least one answer?

asdfryan:
Right now you can only "solve" a puzzle by submitting an answer. Your
suggestion is a good idea, but not sure I have time to implement it before
7 :-/.

I'll create an issue.

PR #410, opened 2021-01-14
Add link back to hunt index in header

asdfryan:
Addresses #409 
Screen Shot 2021-01-14 at 3 00 10 PM


rgossiaux (approved):
Cool, thanks!

Could alternatively put it in the grey navbar at the top.

PR #426, opened 2021-01-15
[bot] Fetch puzzles by tags

akirabaruah:


rawxfish (approved):

chat/management/commands/rundiscordbot.py
```diff
@@ -68,6 +39,29 @@ async def send_puzzles_stuck(message):
...
+            hunt=settings.BOT_ACTIVE_HUNT,
+            tags__in=matched_tags,
+        ).select_related("chat_room")
+    )
+    await send_puzzles(message, puzzles, f"Puzzles tagged with {tags}")
+
+
+async def send_tags(message):
+    tags = await sync_to_async(list)(PuzzleTag.objects.all())
+    embed = discord.Embed(description="`!puzzles tagged <tag> ...` (use `_` for space)")
```
asdfryan, line 60: Why is it necessary to use '_' for space? Can't we just treat tokens 3 onward as a single tag?

akirabaruah:
I'm going to merge this PR for now. We can revert it if it causes problems, but it's more likely that we can just make incremental fixes rather than rewriting the whole thing.

Issue #268, opened 2020-12-23
Shouldn't be able to change the color of a meta tag


rgossiaux:
Right now (in both the old UI and the new React UI) you can change the color of a meta tag by trying to create a new tag with the same name. If the meta is called "Meta", you can create a tag with the name "Meta" and some other color like blue, and it'll change the meta tag's color.

victorchan314:
Will change the behavior to prevent creation of a tag that already exists.

rgossiaux:
Or maybe just prevent changing the color? So if the tag exists, then just
use its existing color

victorchan314:
From a UI perspective it doesn't really make sense to allow creation of an existing tag and silently merge it with the existing tag. Unless you want to allow duplicate tags with multiple colors and keep a separate tag ID for each one, which might be confusing. Thoughts?

rgossiaux:
Nah, we enforce unique names per tag & I don't think we should change that. I think I was thinking that you might be trying to create a tag for something without noticing that the tag already existed and then applying that existing tag could be friendlier. But you're right that that sounds like it could be confusing so I'm cool with whatever you want to do.

victorchan314:
Just kidding, I didn't realize that creating a new tag and adding a tag to a puzzle use the same API. It's not possible to distinguish between the two operations without separating them into different APIs, so I'm just going to implement the silent behavior. Aka creating a tag that exists will change the color of the tag if it is not a meta tag.

PR #427, opened 2021-01-15
Do not change color of a meta tag

victorchan314:
#268

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -340,10 +340,7 @@ class PuzzleTagViewSet(viewsets.ModelViewSet):
                     # the post m2m hook will add tag
                     puzzle.metas.add(meta)
             else:
-                PuzzleTag.objects.filter(
-                    name=tag_name,
-                    hunt=puzzle.hunt,
-                ).update(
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -337,9 +336,16 @@ def create(self, request, **kwargs):
                         },
                         status=400,
                     )
-                # the post m2m hook will add tag
-                puzzle.metas.add(meta)
+                else:
```
asdfryan, line 339: This else is not necessary, no?

victorchan314: No, but I think an explicit else statement is easier to read.

asdfryan: Fair enough.

asdfryan (approved):

asdfryan:
Looks like black is showing diffs, but otherwise LGTM

rgossiaux (changes requested):
Don't actually have any feedback (haven't read it) but since we're code freezing we shouldn't merge this until after the hunt

rgossiaux (approved):
Thanks! (belatedly)

PR #440, opened 2021-01-16
Included embedded urls in all announcements.

asdfryan:


rgossiaux (approved):
Lgtm, let's not deploy this right away though

PR #444, opened 2021-01-16
Rename sheets when puzzles are renamed

### LINKED ISSUE #431: Renaming puzzles should rename associated Google Sheet

rgossiaux:
Fixes #431

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -200,12 +200,24 @@ class PuzzleViewSet(viewsets.ModelViewSet):
                     puzzle.status = data["status"]
                     puzzle.save()
 
-                if puzzle.chat_room and "name" in data and data["name"] != old_name:
-                    puzzle.chat_room.name = data["name"]
-                    puzzle.chat_room.save()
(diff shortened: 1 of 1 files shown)


api/views.py
```diff
@@ -200,14 +200,26 @@ def partial_update(self, request, pk=None, **kwargs):
...
+                if "name" in data and data["name"] != old_name:
+                    if puzzle.chat_room:
+                        puzzle.chat_room.name = data["name"]
+                        puzzle.chat_room.save()
+                        transaction.on_commit(
+                            lambda: chat.tasks.handle_puzzle_rename.delay(
+                                puzzle.id, data["name"]
+                            )
+                        )
+                    if puzzle.sheet and google_api_lib.enabled():
```
asdfryan, line 212: Does this path get called when puzzle gets solved via adding answer?

asdfryan: Oh I see it already gets called. nvm.

asdfryan (approved):

PR #456, opened 2021-01-24
Added puzzle rename announce

jamqwert:
image

### DIFF
--- a/chat/tasks.py
+++ b/chat/tasks.py
@@ -71,10 +71,12 @@ def handle_answer_change(puzzle_id, old_answer, new_answer):
 
 
 @shared_task
-def handle_puzzle_rename(puzzle_id, new_name):
+def handle_puzzle_rename(puzzle_id, old_name, new_name):
     puzzle = Puzzle.objects.get(id=puzzle_id)
     try:
         puzzle.chat_room.handle_puzzle_rename(new_name)
+        msg = f"**{old_name}** has been renamed to **{new_name}**."
+        puzzle.chat_room.send_and_announce_message_with_embedded_urls(msg, puzzle)
     except Exception as e:
         logger.warn(f"Chat operations failed with error: {e}")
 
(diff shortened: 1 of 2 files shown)


asdfryan (approved):

chat/tasks.py
```diff
@@ -71,10 +71,12 @@ def handle_answer_change(puzzle_id, old_answer, new_answer):
 
 
 @shared_task
-def handle_puzzle_rename(puzzle_id, new_name):
+def handle_puzzle_rename(puzzle_id, old_name, new_name):
     puzzle = Puzzle.objects.get(id=puzzle_id)
     try:
         puzzle.chat_room.handle_puzzle_rename(new_name)
+        msg = f"**{old_name}** has been renamed to **{new_name}**."
+        puzzle.chat_room.send_and_announce_message_with_embedded_urls(msg, puzzle)
```
rgossiaux, line 79: The structure here is kind of confusing (not clear without tracing the code path what chat_room.handle_puzzle_rename takes care of vs what we're doing here), wonder if we should move this into `chat_room.handle_puzzle_rename`

asdfryan: Jamie and I actually discussed this offline. I think there are 2 options -- the one you suggested, and keeping this here, and renaming `chat_room.handle_puzzle_rename` to `chat_room.rename_chat` or equivalent such that it's clear that it is only renaming the channels and doing nothing else. I have a slight preference for the latter, but don't feel strongly.

rgossiaux: Yeah that would work too. imo the important thing is that we have all this business logic in one place, and the two options are having them inside the model vs inside the tasks. I don't really feel strongly either.

Now with the benefit of hindsight I wonder if we could make the separation of responsibilities a little clearer. Feels to me like other parts of the app should just notify the chat component about data changes (that a puzzle was created or deleted, an answer edited, etc) and then the specifics of what that actually means (a channel being renamed, a message being sent to the chat channel and/or announcements, etc) are implementation details internal to the chat model. I guess that's mostly how it works right now except for `create_chat_for_puzzle()`, but it's a little confusing since the chat_room model exposes a public API as well as the API of the tasks file.

The other thing is maybe we should change the related name from `puzzle.chat_room` to `puzzle.chat` or something. I think it's a little confusing right now because `puzzle.chat_room` doesn't represent a single room (has voice + video) and is also responsible for handling server-wide things (announcements). Or maybe we can move announcements out of this model somehow, idk.

PR #474, opened 2021-03-31
ignore celery results

### LINKED ISSUE #472: Celery tasks should not be stored in DB
There are O(5k) celery tasks in our DB:
image

I got an email from Heroku:

_The database DATABASE_URL on Heroku app smallboard is approaching its allocated storage capacity.The database contains 7,672 rows. The Hobby-dev plan allows a maximum of 10,000 rows. If the databases exceeds 10,000 rows then INSERT privileges will be revoked, preventing more data from being written. INSERT privileges are automatically reinstated if rows are removed and the database once again complies with the plan limit.To avoid a disruption to your service, migrate the database to a Basic ($9/month) or higher plan:https://hello.heroku.com/upgrade-heroku-postgres#upgrading-with-pg-copy_

_If you are unable to upgrade the database, you should limit or reduce the number of records in it._

We could just pay $9 a month, but I'm fairly certain that we don't need these in our DB.

As Akira alluded to, we should just have Redis manage these instead of storing them in our DB.

rawxfish:
Fixes #472

### DIFF
--- a/google_api_lib/sync_tasks.py
+++ b/google_api_lib/sync_tasks.py
@@ -8,7 +8,7 @@ HUMAN_DRIVE_FOLDER_NAME = "FOLDER FOR HUMANS"
 
 
 @shared_task(base=GoogleApiClientTask, bind=True)
-def get_file_user_emails(self, file_id):
+def get_file_user_emails(self, file_id) -> List[str]:
     response = (
         self.drive_service().files().get(fileId=file_id, fields="permissions").execute()
     )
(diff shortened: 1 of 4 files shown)


asdfryan (approved):

rgossiaux:
Weren't we keeping these for a reason? Maybe getting emails from a drive folder? I thought we had at least one task where we were using the result.

rawxfish:
I think only one function in tasks.py returns anything and it's not used. The stuff where useful things are returned, like getting the emails, is in sync_tasks.py

I removed the return on that one function for clarity and add some typing for the rest.

Empirically, sheet creation is still working for me locally.

rgossiaux (approved):
ah yeah I was thinking of the things in sync_tasks.py. Ok cool, thanks. It looks like these are only useful for getting insights about task statuses/failures but we're not doing that.

PR #478, opened 2021-04-02
put back SECURE_SSL_REDIRECT

rawxfish:
I think we still want all http to redirect to https


rgossiaux (approved):
I think this ended up not actually working (we got an infinite redirect loop) so I turned it off, which is why I deleted the line. But maybe it'll work now with the SECURE_PROXY_SSL_HEADER change.

PR #491, opened 2021-09-21
update packages

rawxfish:


rgossiaux (commented):
Oops sorry I just clobbered this with a dependabot change.

Did you test that everything still works? I see a couple major version changes and one downgrade.

rawxfish:
Nevermind, updating packages broke collectstatic, going close this and do separate PRs for the js updates and python ones to figure out what's going on.

Issue #502, opened 2021-11-04
Update hunt setup documentation


erwa:
Many of the environment variables have changed, there are no instructions for Discord setup, and we could add more info about Heroku setup (adding Postgres and Redis addons, configuring buildpacks (nodejs + python), enabling workers).

erwa:
Resolving this. Follow-up documentation tasks can be addressed in #504 and #445.

PR #503, opened 2021-11-04
Add info on where to create a service account and OAuth 2.0 Client an…

erwa:
…d how to configure the related environment variables accordingly. Also add a pointer to Docker Desktop WSL2 backend.

Part of a bunch of doc updates I plan to do for #502.

### DIFF
--- a/dev-guide.md
+++ b/dev-guide.md
@@ -39,7 +39,9 @@ See below for more details on the `.env` file. You probably want to set at least
 
 ### Docker setup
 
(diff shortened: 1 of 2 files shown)


rgossiaux (approved):
awesome, thanks!

erwa:
Hah, the presubmit failed due to

```
Detect Private Key.......................................................Failed
- hook id: detect-private-key
- exit code: 1

Private key found: new-hunt-setup.md
```

I think it doesn't like the `-----BEGIN PRIVATE KEY-----\n...<long base64-encoded key>...\n-----END PRIVATE KEY-----\n` part I added.

PR #513, opened 2021-11-30
If a tag has the same name as a non-meta puzzle, let it be removed

### LINKED ISSUE #497: "You cannot remove a meta's tag (REDACTED) from itself"
puzzle seems to have a non-meta tag that's the same name as the puzzle itself

rgossiaux:
Fixes #497


erwa (approved):
Nice!

rawxfish (approved):

PR #520, opened 2021-12-14
Query whitelisted emails at authentication time (fixes #500)

### LINKED ISSUE #500: Support adding users / refreshing whitelisted users without restarting app
Currently, smallboard fetches the emails of the users who have access to the `GOOGLE_DRIVE_HUNT_FOLDER_ID` [once on start-up](https://github.com/cardinalitypuzzles/smallboard/blob/7ca1897e8686c900f0e17ee3f99562270c0dc78e/smallboard/settings.py#L244) and sets `SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_EMAILS` equal to this. Oftentimes, you want to add people to your team after the hunt has started, and it would be easier if you could just add the person to the Google Drive and have smallboard automatically pick up the new user without requiring a restart. (Currently, after adding a new person to Google Drive, you need to re-deploy / restart the app.)

erwa:
Before, the list of whitelisted emails was only calculated once at start-up time, and every time you added a user, you would need to restart Cardboard before that user could log in.

Now we just query the list of emails added to the Google Drive as part of the authentication flow. So as soon as you add a user to a Google Drive, they should be able to login to Cardboard.

The Google Drive API to get the folder permissions is pretty fast, so it doesn't noticeably slow down the login. Also, according to https://support.google.com/a/answer/10445916?hl=en, the default Drive API quota is 10,000 calls every 100 seconds, which is plenty for us.

My implementation of `auth_allowed` is based off the default `auth_allowed` implementation here:
* https://github.com/python-social-auth/social-core/blob/911957bc2c0cf0f2e3419e1f1e7caaf25059dc6d/social_core/pipeline/social_auth.py#L12-L14
* https://github.com/python-social-auth/social-core/blob/911957bc2c0cf0f2e3419e1f1e7caaf25059dc6d/social_core/backends/base.py#L148-L159

### DIFF
--- a/cardboard/settings.py
+++ b/cardboard/settings.py
@@ -249,18 +249,16 @@ except KeyError as e:
     )
 
 SOCIAL_AUTH_PIPELINE = (
-    'social_core.pipeline.social_auth.social_details',
-    'social_core.pipeline.social_auth.social_uid',
-
+    "social_core.pipeline.social_auth.social_details",
+    "social_core.pipeline.social_auth.social_uid",
     # custom auth_allowed check in lieu of default one
-    'google_api_lib.sync_tasks.auth_allowed',
-
-    'social_core.pipeline.social_auth.social_user',
-    'social_core.pipeline.user.get_username',
-    'social_core.pipeline.user.create_user',
-    'social_core.pipeline.social_auth.associate_user',
-    'social_core.pipeline.social_auth.load_extra_data',
-    'social_core.pipeline.user.user_details',
+    "google_api_lib.sync_tasks.auth_allowed",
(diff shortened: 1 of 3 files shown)


rgossiaux (approved):
Wow, nice! Impressed how simple this ended up being.

erwa:
My first attempt was to run a periodic task on Celery that updated the email whitelisted every minute or so, but then I found out Django doesn't like it if you update settings at runtime -- apparently they're supposed to be immutable: https://stackoverflow.com/questions/6528723/changing-django-settings-at-runtime

I think this approach is easier and cleaner!

PR #522, opened 2021-12-22
add pre-commit to Pipfile

rawxfish:
Adds pre-commit to Pipfile, so that the stuff that Github runs when we do pull request runs first locally when commit is created.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ celery = "*"
 django-celery-results = "*"
 redis = "*"
 "discord.py" = "*"
+pre-commit = "*"
 
 [dev-packages]
 pre-commit = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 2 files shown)


erwa:
For this to work, you need to run `pre-commit install` manually first in order for the git hook scripts to get installed in `.git/hooks/pre-commit`, right? (per https://pre-commit.com/#3-install-the-git-hook-scripts)

Could we update the dev-guide.md to mention this?

erwa:
I like this pre-commit hook. Thanks for adding it!

rawxfish:
I think this is already in the dev guide: https://github.com/cardinalitypuzzles/cardboard/blob/master/dev-guide.md#set-up-pre-commit-checks

Pipfile
```diff
@@ -18,6 +18,7 @@ celery = "*"
 django-celery-results = "*"
 redis = "*"
 "discord.py" = "*"
+pre-commit = "*"
 
 [dev-packages]
 pre-commit = "*"
```
rgossiaux, line 24: Isn't this already in the dev-packages section below? I think if you run `pipenv install --dev` it should be installed.

rawxfish: oh you're right, I forgot the --dev, thanks!

PR #525, opened 2021-12-25
adds toggle for light and dark mode

rawxfish:
Light:

image

Dark:

image

### DIFF
--- a/cardboard/static/style.css
+++ b/cardboard/static/style.css
@@ -27,3 +27,17 @@ body {
 .table td {
     vertical-align: baseline;
 }
+
+/* Dark/Light Mode Toggle Overrides */
+.btn label.switch-on, .btn label.switch-off {
+    margin-bottom: 0px;
+    border: 0;
+    border-radius: 0;
+}
+
+
+/* Dark Mode Overrides */
+
+.bootstrap-dark a {
+    color: #007bff;
+}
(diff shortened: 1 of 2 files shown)


rgossiaux:
Hmm I feel like we might need to change a lot of the other colors with dark mode, this feels kinda painful to look at it (not that the normal UI is that pretty to begin with...)

erwa (approved):
Maybe ask some of the folks who requested dark mode to take a look at the screenshot and see what they think?

(I'm fine with what you've done so far; we can always tweak the colors further based on feedback.)

rawxfish:
Hmm I didn't want to spend too much time tweaking colors, so I just went with the first templates I could find lol 

Alternatively, we can use some of the themes from https://bootswatch.com/ if you guys prefer those?

rgossiaux:
> Hmm I didn't want to spend too much time tweaking colors, so I just went with the first templates I could find lol
> 
> Alternatively, we can use some of the themes from https://bootswatch.com/ if you guys prefer those?

Yeah, I guess we can try a couple of those and see if any of them look better?

rawxfish:
I played around with the dark scenes from that... I think all of them would also require a lot of manual overrides. For now, I think we should just go with this, since it at least adds the toggle and makes it easier to do the overrides via the .bootstrap-dark class

Issue #437, opened 2021-01-16
Persist UI selections per user


akirabaruah:
For example, after selecting the "Unsolved" toggle and refreshing the Hunt puzzles page, the selection reverts to "All" (default). It would be nice if it stayed the same for the duration of my session.

PR #528, opened 2021-12-28
Persist solve state filter

pcvera:
This PR adds persists the 'solve state' filter at the top of the page to local storage. The state that controlled that behavior had to be refactored into redux state to take advantage of a library that made this simple in redux

[see issue #437](https://github.com/cardinalitypuzzles/cardboard/issues/437)


hunts/src/store.js
```diff
@@ -12,5 +18,13 @@ export default configureStore({
     puzzles: puzzlesReducer,
     hunt: huntReducer,
     tagFilter: tagFilterReducer,
+    filter: filterReducer,
```
erwa, line 21: Is it easy to combine tagFilter and filter into one file? Logically they are all different kinds of puzzle filters. It seems we have 3 kinds right now: text filter, solved filter, and tag filter. I think it would be a little cleaner to have a single filterReducer that applies all 3.

(It would also be nice for localStorage to remember the last-applied tag filter, too.)

erwa (commented):
Nice improvement. My main question is if the "filter" you added can be combined with the existing "tagFilter". Logically they seem to do similar things (filter what puzzles are shown).

hunts/src/store.js
```diff
@@ -12,5 +18,13 @@ export default configureStore({
...
   },
+  middleware: (getDefaultMiddlewares) => [
+    ...getDefaultMiddlewares(),
+    save({
+      states: ["filter.solveStateFilter"],
```
erwa, line 26: I see you are saving just the solveStateFilter here. What do you think about saving the getSolveStateFilter and tagFilter too?

hunts/src/filterSlice.js
```diff
@@ -0,0 +1,36 @@
+import { createSlice, createSelector } from "@reduxjs/toolkit";
+
+export const SOLVE_STATE_FILTER_OPTIONS = {
+  ALL: 0,
+  UNSOLVED: 1, // Excludes solved feeders and everything below them.
```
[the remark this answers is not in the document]
rgossiaux: I think "solved metas and everything below them", or "solved metas and all their feeders"

rgossiaux (approved):
Nice, thanks for refactoring the filter component out too. LGTM though I also like Anthony's suggestion about saving the other filters while we're at it.

hunts/src/store.js
pcvera: I added the tag filter to the persisted data. 

I tried adding the text filter too, but it felt really unnatural to open the page and have that value set, so I omitted it.

hunts/src/solveStateFilter.js
```diff
@@ -58,3 +60,73 @@ export function filterSolvedPuzzlesfn(rows, id, filterValue) {
...
+                  )
+                );
+              } else {
+                dispatch(
+                  updateSolveStateFilter(SOLVE_STATE_FILTER_OPTIONS.UNSOLVED)
+                );
+              }
+            }}
+          ></input>
+          Include unsolved w/ solved metas
```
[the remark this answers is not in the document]
pcvera: I don't recall, this string has been in the codebase since last year. Probably space?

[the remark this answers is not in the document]
erwa: I'm on my work laptop so not able to easily run Cardboard right now to check, but if it looks ok spelled out, my vote is for "with".

hunts/src/store.js
erwa: That's fair! Thanks for adding the tag filter.

hunts/src/solveStateFilter.js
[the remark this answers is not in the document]
pcvera: done

hunts/src/store.js
pcvera: done

erwa:
Nice! Thanks for the updates Peter. Looks good.

PR #530, opened 2022-01-02
Random unused import fixes

rgossiaux:


rawxfish (approved):

Pipfile
```diff
@@ -18,6 +18,8 @@ celery = "*"
 django-celery-results = "*"
 redis = "*"
 "discord.py" = "*"
+django-stubs = "*"
+celery-types = "*"
```
erwa, line 22: so after adding these 2 dependencies, what did you run to find the unused imports and local variables? is it possible to add this as a precommit check?

erwa (approved):

erwa:
looks like the pre-submit tests failed due to some dependency conflicts presumably due to the new deps you added. I guess this is what you were referring to in the Discord chat about dependency issues?

Pipfile
rgossiaux: They were surfaced by the pyright LSP (I use neovim). I think VSCode will show these too but since I don't use it I can't say any more.

rgossiaux:
> looks like the pre-submit tests failed due to some dependency conflicts presumably due to the new deps you added. I guess this is what you were referring to in the Discord chat about dependency issues?

Actually my problem in Discord was that I couldn't lock the dependencies at all even with a clean repo. But yeah, this change also seems to be introducing some dependency issues. Hadn't had time to look into it earlier but I'll give it a stab now.

rgossiaux:
Actually not going to add this type stub library... it's optimized for mypy and editor integrations use pyright so it doesn't fix everything. There's a fork that removes the mypy dependency https://github.com/sbdchd/django-types but you have to do ugly some manual work in the codebase for it to pick up things like related models, don't want to do that if I'm going to be the only one using this.

For now just going to merge in the random import fixes

PR #531, opened 2022-01-03
change tags name to case insensitive class

rawxfish:
Hmm I'm concerned that CICharField will break tests, since it is a postgres-specific field and our tests might be using sqlite, but making PR to be sure


rawxfish:
nvm, looks like django is doing something smart for postgres specific fields when sqlite is used.

rawxfish:
jk, looks like the migration messes up if there is existing data that is not unique after case insensitivity. I will fix it.

rgossiaux (approved):
LGTM, thanks!

puzzles/puzzle_tag.py
```diff
@@ -22,7 +23,7 @@ class PuzzleTag(models.Model):
     hunt = models.ForeignKey(
         "hunts.Hunt", on_delete=models.CASCADE, related_name="puzzle_tags"
     )
-    name = models.CharField(max_length=100)
+    name = CICharField(max_length=100)
```
erwa, line 26: what's the motivation for changing from a more general CharField to a Postgres-specific CICharField? do we already have Postgres-specific stuff in our code?

(sorry I am late to review this PR)

erwa: I guess you're trying to use CICharField to enforce we only have one of Logic/LOGIC/LoGiC stored in the DB.

I kinda like the flexibility of not having DB-specific fields so we're less likely to have issues if we change DBs in the future (or for local testing). I would be more in favor of standardizing the tag names in frontend validation rather than relying on a Postgres-specific CICharField to do the enforcement. We could use a clean() method like here: https://stackoverflow.com/a/9498470/1128392. Though the tradeoff is we'd need to pick a standardized tag format like ALL_UPPERCASE or all_lowercase (I'd vote for ALL_UPPERCASE). I guess meta tags would also become ALL_UPPERCASE in this case?

What you've done now is fine I guess; I think in practice if Postgres and SQLite both work with CICharField, then using it shouldn't be a big impediment to using other DBs.

rgossiaux: I think I'd rather do it API-side than on the frontend. Personally I actually don't like the ALL_UPPERCASE tag names (harder to read) and was thinking about sending a PR to change the default ones to Sentence case.

With that said I've shifted my position on this over time & I actually don't mind using a few Postgres-specific features now. I feel like that's the part of the stack that's the least likely to change.

erwa: Yeah, I guess going with the user-inputted casing is the most user-friendly. And keeping the meta tag names Sentence Cased seems most natural since the Meta puzzle titles are generally Sentence Cased.

PR #533, opened 2022-01-04
pipenv => poetry

rgossiaux:
I'm having problems with pipenv on my system and poetry Just Works. I think it's also more popular than pipenv these days for new projects, so it shouldn't be a bad move.


rawxfish (approved):

erwa (commented):
Nice, thanks for updating the Docker script and the docs.

Should we add `requirements.txt` to `.gitignore` so we don't accidentally check in a requirements.txt which might upset Poetry?

PR #534, opened 2022-01-06
Remove the border/background around the icons

rgossiaux:
The UI needs more work than this (especially the answer cell) but I think it's a step in the right direction. Looks a lot cleaner to me without twenty pills popping out the screen at once.

Before: 
image

After:
image

### DIFF
--- a/cardboard/static/style.css
+++ b/cardboard/static/style.css
@@ -25,7 +25,7 @@ body {
 }
 
 .table td {
-    vertical-align: baseline;
+    vertical-align: middle;
 }
 
 /* Dark/Light Mode Toggle Overrides */
--- a/hunts/src/AnswerCell.js
+++ b/hunts/src/AnswerCell.js
@@ -1,12 +1,10 @@
-import Badge from "react-bootstrap/Badge";
 import Button from "react-bootstrap/Button";
 import React from "react";
-import { updatePuzzle } from "./puzzlesSlice";
-import { useSelector, useDispatch } from "react-redux";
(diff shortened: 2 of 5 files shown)


rawxfish (approved):

erwa (approved):
Looks a bit nicer to me too.

erwa:
Er, wait, actually I think I prefer the pills b/c it makes the buttons look more differentiated from the text. After this change, especially the answer column looks a bit strange to me: the `ANSWER <icon> <icon>` almost looks to me like the answer has 2 emojis in it.

Uh, this might just be a flavors of ice cream thing, so I'm not gonna sweat it. Feel free to merge it if you and others like it.

erwa:
is the "After" screenshot with the vertical-align: middle? if not, can you include a screenshot with that change?

rgossiaux:
> Er, wait, actually I think I prefer the pills b/c it makes the buttons look more differentiated from the text. After this change, especially the answer column looks a bit strange to me: the `ANSWER <icon> <icon>` almost looks to me like the answer has 2 emojis in it.
> 
> Uh, this might just be a flavors of ice cream thing, so I'm not gonna sweat it. Feel free to merge it if you and others like it.

Yeah I think the Answer column looks pretty bad still, though I also think it looks bad before this change. It definitely still needs work. Personally I think it's still in the right direction but YMMV.

> is the "After" screenshot with the vertical-align: middle? if not, can you include a screenshot with that change?

Yeah, it is. It's a pretty small difference.

PR #535, opened 2022-01-06
load .env file in celery workers

rawxfish:
Pipenv would automatically load .env files, so it didn't matter if the celery workers didn't. Celery stopped working in local development after switching to poetry, which ignores .env files. It didn't break prod, since the environment variables are set in the Heroku apps instead of through .envs.

### DIFF
--- a/cardboard/celery.py
+++ b/cardboard/celery.py
@@ -1,9 +1,15 @@
 import os
+import dotenv
 
 from celery import Celery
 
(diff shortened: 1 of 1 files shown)


rgossiaux (approved):
Aha interesting. The Docker config also loads the .env so it's not a problem there. Can also do `source env && [your command]` but i don't mind merging this if it makes your life easier

PR #536, opened 2022-01-06
Change puzzle/sheet/chat links to a softer color

rgossiaux:
Trying to stick with Bootstrap colors for now. I think this makes them less prominent/so they don't compete for attention with the big answer/solve buttons. Still underlines & changes color on hover so it still looks like a link. 

Old: image

This version (.text-info): 
image

Alternate (text-secondary):
image

Alternate (text-dark); I don't like this one since it doesn't look like a link: 
image

### DIFF
--- a/hunts/src/HuntViewMain.js
+++ b/hunts/src/HuntViewMain.js
@@ -64,7 +64,7 @@ const TABLE_COLUMNS = [
     Header: "Puzzle",
     accessor: "url",
     Cell: ({ row, value }) => (
-      <a href={value} target="_blank">
+      <a href={value} target="_blank" className="text-info">
         Puzzle
       </a>
     ),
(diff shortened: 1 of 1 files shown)


rawxfish (approved):
I like the text-info version the best too

erwa (approved):
I slightly prefer the original default blue link color (as I'm just used to links being blue) but I'm fine with this change if you guys prefer the greener color.

Also, something else I noticed from the screenshot that might be a nice visual fix is to make the SOLVING/STUCK/etc. dropdown always the same width so visually the Status column looks uniform. Right now, the dropdown box width varies so there is a jagged right edge when you look down that column.

rgossiaux:
> I slightly prefer the original default blue link color (as I'm just used to links being blue) but I'm fine with this change if you guys prefer the greener color.

I wouldn't say I *love* it but I do think having the three blue links right next to the two blue buttons is too much blue. Maybe changing the button color is an alternate solution, I haven't played with that yet.

> Also, something else I noticed from the screenshot that might be a nice visual fix is to make the SOLVING/SOLVED/STUCK/etc. dropdown always the same width so visually the Status column looks uniform. Right now, the dropdown box width varies so there is a jagged right edge when you look down that column.

Oh yeah that's a good idea.

rgossiaux:
Closing this since we're switching them to icons instead (much better lol)

PR #537, opened 2022-01-06
Pin dependencies in pyproject.toml

rgossiaux:
The ^ notation means any version that matches the leftmost non-zero number will be accepted. This way we don't accidentally upgrade a breaking change and break cardboard.


rawxfish (approved):

pyproject.toml
```diff
@@ -7,20 +7,20 @@ license = "MIT"
...
+djangorestframework = "^3.13.1"
 disco-py = ">=0.0.14rc1"
-celery = "*"
-django-celery-results = "*"
-redis = "*"
-"discord.py" = "*"
+celery = "^5.2.3"
+django-celery-results = "^2.2.0"
+redis = "^4.1.0"
+"discord.py" = "^1.7.3"
```
erwa, line 23: Can you use an asterisk for part of the version, e.g.: `^4.*`? Might be nice to automatically pick up minor and patch version bumps. (I think this would also reduce the number of Dependabot security version bump PRs we get.) Though using floating minor and patch versions might also cause the poetry.lock file to change frequently and cause unnecessary noise in PRs. What do you guys think?

rgossiaux: That's what the ^ does: https://python-poetry.org/docs/dependency-specification/ So eg `^4.0.1` means any 4.x.y version that's  >=4.0.1. The only difference is that it excludes `4.0.0` specifically. Both probably work for us I think.

erwa: oooh got it now
Thanks for explaining.

erwa (approved):

PR #538, opened 2022-01-07
First stab at per-hunt settings

rgossiaux:
This commit
  * adds a HuntSettings model which contains all the settings for a hunt.
  * initializes HuntSettings for existing Hunts using the current values for environmental variables in your environment
  * creates a new HuntSettings object when a Hunt is created
  * uses the HuntSettings preferentially for Google-related tasks (the human Drive folder, creating new sheets), falling back to env variables if not set
  * exposes the HuntSettings in the Django admin

Can follow-up with a change to use HuntSettings for Discord-related tasks as well.

For supporting multiple independent users, after that we still need to:
  * expose a way to set/update these in the UI
  * do all the permissions stuff

which there is definitely not time for this year.

Concretely this will at least let us change these settings during hunt if for some reason we need to, by using the django admin.


hunts/models.py
```diff
@@ -133,3 +137,59 @@ def get_minutes_per_solve(self, recent=False):
...
+    # Google-specific settings
+    #
+
+    # The id of your Google Drive folder
+    # This should be part of the URL (https://drive.google.com/drive/folders/<folder_id>)
+    google_drive_folder_id = models.CharField(max_length=128, blank=True)
+
+    # The id of your Google Sheets template file
+    # This should be part of the url (https://docs.google.com/spreadsheets/d/<sheet_id>)
+    google_sheets_template_file_id = models.CharField(max_length=128, blank=True)
```
rawxfish, line 159: Right now we're just using google_drive_folder_id for figuring out who can log in. New sheets are created in the same folder that the template is in. There is technically nothing enforcing that these are the same folder, so that's another future TODO I think.

hunts/models.py
```diff
@@ -133,3 +137,59 @@ def get_minutes_per_solve(self, recent=False):
...
+
+    # The id of the Discord channel to make puzzle announcements in
+    # This channel can get noisy and is recommended to be its own separate channel
+    # This ID can be found by enabling Developer Mode in Discord and right-clicking on the channel
+    discord_puzzle_announcements_channel_id = models.CharField(
+        max_length=128, blank=True
+    )
+
+    # The category name to create all Discord text channels in
+    discord_text_category = models.CharField(
```
rawxfish, line 180: We will not need `discord_text_category` and `discord_voice_category` anymore after we switch over to organizing channels under their meta, right?

rawxfish (approved):
Nice!!

hunts/models.py
rgossiaux: Yeah that's a good point, added a TODO.

hunts/models.py
rgossiaux: Right, these fields might end up being pretty temporary.

hunts/models.py
erwa: Oh wow, I never realized that. Totally thought `GOOGLE_DRIVE_HUNT_FOLDER_ID` was where new puzzle sheets were created

erwa: Filed https://github.com/cardinalitypuzzles/cardboard/issues/541 for this. Can we updated the TODO to TODO(#541)?

erwa (approved):
This is awesome! Looks clean

hunts/models.py
rgossiaux: Thanks, done.

PR #540, opened 2022-01-07
add option to switch discord links from https:// to discord://

rawxfish:
So that those with app installed can click to open discord and change to the proper text channel.

image

### DIFF
--- a/hunts/src/chatSlice.js
+++ b/hunts/src/chatSlice.js
@@ -1,0 +1,22 @@
+import { createSlice, createSelector } from "@reduxjs/toolkit";
+
+export const CHAT_VERSION_OPTIONS = {
+  APP: 0,
+  WEB: 1,
+};
+
+export const chatSlice = createSlice({
+  name: "chat",
+  initialState: {
+    version: CHAT_VERSION_OPTIONS.APP,
+  },
+  reducers: {
+    updateChatVersion: (state, action) => {
+      state.version = action.payload;
+    },
+  },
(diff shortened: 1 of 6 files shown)


hunts/src/chatSlice.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const CHAT_VERSION_OPTIONS = {
+  APP: 0,
+  WEB: 1,
+};
+
+export const chatSlice = createSlice({
+  name: "chat",
+  initialState: {
+    version: CHAT_VERSION_OPTIONS.APP,
```
erwa, line 11: What happens if you don't have the Discord app installed and click a `discord://` link? If the link won't work in that case, I think we should make the default WEB for a better out-of-the-box experience. If it redirects you to install the Discord app, then I think it's ok to make APP the default.

rawxfish: I actually prefer app be the default, because (1) there seems to be more people who use app on our team and (2) this feature improves discord navigation for app users but they might miss it if the default is web because it's not so obvious that there's something better. For ppl without the app, it is very obvious that they need to switch modes since it just google searchers the link

erwa: Ok, that's fair. I guess we are not selling this product to potential customers where I think it'd be more important that everything just works out-of-the-box without additional configuration. We can optimize for the majority user on our team then.

erwa (approved):

PR #543, opened 2022-01-08
improve dark mode colors

pcvera:
I don't know much about color but this feels better to me?

Before
image

After
Light mode (unchanged)
image

Dark mode (improved)
image


rawxfish (approved):
woo nice

PR #556, opened 2022-01-10
update tools page

rawxfish:
Added a few more tools and updated last year's web to this years

### DIFF
--- a/cardboard/templates/tools.html
+++ b/cardboard/templates/tools.html
@@ -159,7 +159,6 @@
                         <li><a href="http://puzzles.mit.edu/tools.html">MITMH Puzzles Tools Page</a></li>
                         <li><a href='https://thetech.com/issues/140/43/pdf'>MIT Newspaper</a></li>
                         <li><a href='https://docs.google.com/spreadsheets/d/1vxO5hI2Uqr8rB5QOA3u0AOQYj6CSBeCyPVAylaZZvbg/edit?usp=sharing'>Useful Links</a></li>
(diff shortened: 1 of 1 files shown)


cardboard/templates/tools.html
```diff
@@ -140,25 +151,15 @@ <h1>Tools and References</h1>
...
-
         <div class="card" >
             <div class="card-body">
                 <div class="card-title"><h2>MITMH Specific</h2></div>
                 <div class="card-text">
                     <ul>
+                        <li><a href="http://puzzles.mit.edu/tools.html">MITMH Puzzles Tools Page</a></li>
                         <li><a href='https://thetech.com/issues/140/43/pdf'>MIT Newspaper</a></li>
                         <li><a href='https://docs.google.com/spreadsheets/d/1vxO5hI2Uqr8rB5QOA3u0AOQYj6CSBeCyPVAylaZZvbg/edit?usp=sharing'>Useful Links</a></li>
+                        <li><a href='https://docs.google.com/drawings/d/1n0DcJpA4dCd63dUnqG2TFmd006FtBJEMrVVfDUPiVt4/edit'>Cardinality connections web</a></li>
```
pcvera, line 162: did you mean to add this?

rawxfish: I was actually sorta torn whether I should include it, what do yall think?

pcvera: If we're expecting non-Cardinality people to use Cardboard, seems kinda weird that they'd be able to see the link at all. 

We could add it to a future Cardinality team website?

rgossiaux: I think it's okay to leave out since we have other ways of distributing it (Discord or email) and I feel like people are going to look for it there anyway. IMO we should feel comfortable putting stuff on here if we think it's useful for us even if it's not useful for other people, though.

rgossiaux (approved):

PR #557, opened 2022-01-10
clean up channels 30 minutes after puzzle solve

rawxfish:
Delete voice channels and text channels that contain no non-bot messages 30 minutes after a puzzle is marked as SOLVED.

For puzzles with multiple answers, the cleanup is avoided by marking the puzzle as SOLVING after putting in the first answer.

### DIFF
--- a/chat/tasks.py
+++ b/chat/tasks.py
@@ -1,6 +1,9 @@
 from cardboard.settings import TaskPriority
 from celery import shared_task
 import logging
+from django.db import transaction
+from datetime import datetime, timedelta
+
 
 from puzzles.models import Puzzle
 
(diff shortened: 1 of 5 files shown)


chat/tasks.py
```diff
@@ -26,18 +37,19 @@ def handle_puzzle_solved(puzzle_id, answer_text):
         msg = f"{puzzle.name} has been solved with {answer_text}!"
         puzzle.chat_room.send_and_announce_message_with_embedded_urls(msg, puzzle)
     except Exception as e:
-        logger.warn(f"Chat operations failed with error: {e}")
+        logger.warn(f"handle_puzzle_solved failed with error: {e}")
```
rgossiaux, line 40: thank you for fixing these 🙌

chat/tasks.py
```diff
@@ -15,7 +15,18 @@ def create_chat_for_puzzle(puzzle_id):
...
         puzzle.chat_room.send_and_announce_message_with_embedded_urls(msg, puzzle)
     except Exception as e:
-        logger.warn(f"Chat operations failed with error: {e}")
+        logger.warn(f"create_chat_for_puzzle failed with error: {e}")
+
+
+@shared_task
+def cleanup_puzzle_channels(puzzle_id):
+    puzzle = Puzzle.objects.get(id=puzzle_id)
+    # check that puzzle wasn't unsolved between when task was queued and now
```
rgossiaux, line 24: It's pretty edge case-y, but if a puzzle was solved, unsolved, and re-solved within 30m, this could delete the channels sooner than expected. Should we maybe add an explicit solved timestamp to the Puzzle model? Looks like it might make the stats slightly more robust too: https://github.com/cardinalitypuzzles/cardboard/blob/195dea2eb57eddcd1bfc9da3b02d8788a8922844/hunts/chart_utils.py#L43

rgossiaux (commented):
This is awesome, thank you!!!

chat/tasks.py
rawxfish: Could https://github.com/cardinalitypuzzles/cardboard/blob/master/puzzles/models.py#L164 work for both? I guess it's not really more robust than what's in chart_utils.py right now.

rgossiaux: Yeah sounds good, fine w/ me if you want to just use that for now. Introducing a separate solve timestamp has its own downsides since now we have to keep it accurate.

rgossiaux (approved):
woo!! hopefully this will mean we don't have to keep muting so many new archive categories this year 😛

PR #562, opened 2022-01-11
Table column width fixes

rgossiaux:
* Force the links to always appear on one line (previously they break onto two or three lines as the screen shrinks)
* Set some basic proportions for the columns widths (tested using last year's hunts and a new hunt)

Last year's MH on my full width laptop screen: 
image

Empty hunt on full width laptop screen:
image

As the screen width goes down this matters less and less because the table layout is constrained by wide elements (long answers/tags/words)


rawxfish (approved):
nice

PR #564, opened 2022-01-11
remove maxims and add section for additional sheet templates

rawxfish:
image


bcai2 (approved):
LGTM, thanks for editing the links!

rgossiaux (approved):
I like the Benji maxims lol, hope we can keep them alive somehow still

PR #568, opened 2022-01-12
Use the Drive folder ID for storing puzzle files instead of the same directory as the template

### LINKED ISSUE #541: Create new puzzle sheets in the hunt drive folder rather than in the folder that the template sheet is in
Currently, we create new puzzle sheets in the same folder that the template sheet is in (by copying the template): https://github.com/cardinalitypuzzles/cardboard/blob/b9e212d848277be3ec038817f580b0db4556cd90/google_api_lib/tasks.py#L18-L31

It makes more sense to create the new puzzle sheets in the hunt drive folder (either GOOGLE_DRIVE_HUNT_FOLDER_ID or the per-hunt drive folder location once #538 is merged).

rgossiaux:
Fixes #541

This lets us keep the template in a different location from the puzzle files, which is useful
if we're potentially going to make a bunch of copies of the template in advance and need to
distinguish those from puzzle sheets.

### DIFF
--- a/google_api_lib/tasks.py
+++ b/google_api_lib/tasks.py
@@ -29,6 +29,17 @@ def create_google_sheets_helper(self, name, template_file_id) -> dict:
     return file
 
 
+@shared_task(base=GoogleApiClientTask, bind=True)
+def move_drive_file(self, file_id, destination_folder_id) -> None:
+    file = self.drive_service().files().get(fileId=file_id, fields="parents").execute()
+
+    self.drive_service().files().update(
+        fileId=file_id,
+        addParents=destination_folder_id,
+        removeParents=",".join(file["parents"]),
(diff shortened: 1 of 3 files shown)


rawxfish (approved):

erwa:
Thanks for fixing this.

Just curious, if the template happens to be in the same folder as the Drive folder, does the code still work (no error returned by the Google APIs)?

rgossiaux:
> Just curious, if the template happens to be in the same folder as the Drive folder, does the code still work (no error returned by the Google APIs)?

Yep, still works. This task is also async from the sheet creation one so even if something does go wrong for some reason, it won't interfere with sheet creation.

erwa:
Thanks for confirming!

PR #569, opened 2022-01-12
Support world-readable hunt folder when checking authentication (fixes #561)

### LINKED ISSUE #561: Better error handling for Google API
Apparently if you share the Drive folder with everybody, the response [here](https://github.com/cardinalitypuzzles/cardboard/blob/4f448f01347f48674e48264712da8ef7d405b9a1/google_api_lib/sync_tasks.py#L35) may not include any email addresses. We should handle that. More generally we should try to generate good error messages for these API calls if we can.

erwa:


google_api_lib/sync_tasks.py
```diff
@@ -18,12 +19,13 @@ def auth_allowed(backend, details, response, *args, **kwargs):
...
-    if settings.GOOGLE_API_AUTHN_INFO and email not in get_file_user_emails(
-        settings.GOOGLE_DRIVE_HUNT_FOLDER_ID
-    ):
-        raise AuthForbidden(backend)
+    # Allow all emails if Google Drive integration is not set up
+    # or folder is world readable. Otherwise, only allow emails
+    # added to the Google Drive folder.
+    if settings.GOOGLE_API_AUTHN_INFO:
+        user_emails = get_file_user_emails(settings.GOOGLE_DRIVE_HUNT_FOLDER_ID)
+        if user_emails and email not in user_emails:
```
rgossiaux, line 27: Hmmhmm should we at least log a warning or error? The other thing is that this doesn't distinguish between the empty list and None, maybe we should check `if user_emails is not None` instead

erwa: I added a log.warn() in my update, though when AuthForbidden is raised, that also already gets logged along with a stack trace.

I also updated to check `user_emails is not None`. In practice, I don't think `user_emails` will ever be an empty list because a folder always has an owner, so I think the list will always have at least one email. I update the docs for get_file_user_emails to explain the behavior (either a list of emails or None if the file is world-readable).

rgossiaux: >In practice, I don't think user_emails will ever be an empty list because a folder always has an owner, so I think the list will always have at least one email.

Oh yeah, you're right, lol. Thanks.

>I added a log.warn() in my update, though when AuthForbidden is raised, that also already gets logged along with a stack trace.

I meant log a warning if `user_emails` is None; it might be unexpected that the Cardboard instance is de facto public and that anyone with the URL can log in.

rgossiaux (approved):

google_api_lib/sync_tasks.py
erwa: Ooh, gotcha. I updated the log message to warn when the Google Drive folder is world-readable and thus anyone can log in to the Cardboard instance.

rgossiaux:
Actually I just remembered that if you do this (link-share your drive folder) then everyone shows up as anonymous animals. Maybe we shouldn't support this at all & force users to add people to the Drive folder.

erwa:
> Actually I just remembered that if you do this (link-share your drive folder) then everyone shows up as anonymous animals. Maybe we shouldn't support this at all & force users to add people to the Drive folder.

This sounds like a policy decision each team can make. I like the flexibility of allowing a team of making everything world-readable if they don't want to bother adding each person individually, even if personally I think it's a better experience if you add all team members individually so they show up with names instead of as anonymous.

rgossiaux:
That's true, but people might not realize that it'll happen. Personally I don't mind taking an opinionated stance if it's going to give people a better experience. But we can leave it how it is.

In my dream world Cardboard just creates the Drive folder when you create a new hunt & it manages the permissions for you, making this a non-issue. Maybe some day...

erwa:
Yep, was thinking the same thing. In the future, if Cardboard manages all permissions (you define users in Cardboard and Cardboard goes and creates the Google Drive folder and sets up the permissions properly), this isn't an issue. I think we'll probably have to move to this model to properly support multi-tenancy since I don't think it makes sense for Cardboard to query all hunts during login to figure out which hunts a user has access to.

PR #570, opened 2022-01-12
Fix the backsolved tag/sheet renaming interaction

rgossiaux:
I broke this when I changed the casing of the default tags. Making this case-insensitive.

### DIFF
--- a/api/tests.py
+++ b/api/tests.py
@@ -374,3 +374,61 @@ class SheetTests(CardboardTestCase, TransactionTestCase):
             rename_sheet.assert_called_with(
                 sheet_url=google_api_lib.tests.TEST_SHEET, name=f"{puzzle.name}"
             )
+
+    @patch("google_api_lib.tasks.rename_sheet.delay")
+    @patch(
+        "google_api_lib.tasks.create_google_sheets_helper",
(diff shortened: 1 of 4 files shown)


rgossiaux:
Looks like the tests are broken here because GH actions are using sqlite so I made a PR to change that over.

rawxfish (approved):

erwa:
> Looks like the tests are broken here because GH actions are using sqlite so I made a PR to change that over.

I'm guessing this is the failed run?: https://github.com/cardinalitypuzzles/cardboard/runs/4786213594?check_suite_focus=true

```
AssertionError: expected call not found.
Expected: delay(sheet_url='testsheet.com', name='[BACKSOLVED: ANS, ANS2] Test')
Actual: delay(sheet_url='testsheet.com', name='[SOLVED: ANS, ANS2] Test')
```

Do we know why it failed with sqlite but passed with Postgres? This doesn't involve the case-insensitive field so I'm surprised at the difference in behavior.

erwa:
> This doesn't involve the case-insensitive field so I'm surprised at the difference in behavior.

Oh nvm, I just looked at the code and you are testing case-insensitive updates. That explains why

erwa (approved):
Nice. Thanks for adding tests!

PR #575, opened 2022-01-13
Support copying existing files from a Drive folder

rgossiaux:
I tested this manually with the folder not set, with the folder set but empty, and with the folder set and with files in it. Seems to work.


hunts/migrations/0009_huntsettings_google_sheets_template_folder_id_and_more.py
```diff
@@ -0,0 +1,27 @@
...
+        ("hunts", "0008_huntsettings_backfill"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="huntsettings",
+            name="google_sheets_template_folder_id",
+            field=models.CharField(blank=True, max_length=128),
+        ),
+        migrations.AlterField(
```
rgossiaux, line 20: Oops this is my bad, I changed this earlier but never generated a migration for it. Looks like it's also included in #573

rawxfish (approved):

google_api_lib/tasks.py
```diff
@@ -55,41 +85,57 @@ def transfer_ownership(self, file, template_file_id) -> None:
...
-    transfer_ownership.delay(response, template_file_id)
-
-    if destination_folder_id:
-        move_drive_file.delay(
-            file_id=response["id"], destination_folder_id=destination_folder_id
+    with transaction.atomic():
+        puzzle = Puzzle.objects.select_related("hunt__settings").get(pk=puzzle_id)
+        template_file_id = (
+            puzzle.hunt.settings.google_sheets_template_file_id
+            or settings.GOOGLE_SHEETS_TEMPLATE_FILE_ID
```
erwa, line 92: can we add mention of this new var to the new-hunt-setup.md?

erwa (commented):
Looks good. One minor comment: let's add the new GOOGLE_SHEETS_TEMPLATE_FILE_ID to the docs.

rgossiaux:
Oops, good catch. #583 to update the documentation.

google_api_lib/tasks.py
```diff
@@ -55,41 +85,57 @@ def transfer_ownership(self, file, template_file_id) -> None:
...
-    puzzle.save()
-
-    transfer_ownership.delay(response, template_file_id)
-
-    if destination_folder_id:
-        move_drive_file.delay(
-            file_id=response["id"], destination_folder_id=destination_folder_id
+    with transaction.atomic():
+        puzzle = Puzzle.objects.select_related("hunt__settings").get(pk=puzzle_id)
+        template_file_id = (
```
erwa, line 90: nit: `template_file_id` is only used in the `else` block below. You could move this assignment inside the `else` block too to keep the declaration/assignment closer to when the var is use.

erwa: oh wait, no, nvm, you use it in the `post_create_tasks` function

PR #582, opened 2022-01-13
fix backfill migrations

rawxfish:
#573 broke this migration since it removed DISCORD_GUILD_ID from settings.

We can wait until after the hunt to merge this in.


rgossiaux (approved):
Thanks!

PR #596, opened 2022-01-17
Add blank=True to a bunch of fields for the admin

### LINKED ISSUE #594: Fix some of the admin pages having non-optional fields
A bunch of fields on the Puzzle model, for example, should be made optional

rgossiaux:
Fixes #594


rawxfish:
shouldn't there be migration files too?

puzzles/models.py
```diff
@@ -62,20 +62,23 @@ class Puzzle(models.Model):
...
+    tags = models.ManyToManyField(PuzzleTag, related_name="puzzles", blank=True)
 
-    metas = models.ManyToManyField("self", symmetrical=False, related_name="feeders")
+    metas = models.ManyToManyField(
+        "self", symmetrical=False, related_name="feeders", blank=True
+    )
 
     is_meta = models.BooleanField(default=False)
 
     active_users = models.ManyToManyField(
```
erwa, line 76: not related to your change, but what was active_users for again?

erwa (approved):
LGTM

puzzles/models.py
rgossiaux: We're not actually using it, lol. I think the idea was to keep track of who's working on a puzzle at a time. We should probably go through and do some cleanup of the stuff we never ended up using or that we deprecated.

rawxfish (approved):

puzzles/models.py
erwa: lol I already forgot I added this feature back in #80...
yeah, doesn't seem like this is used anymore, we can probably just get rid of it

erwa: Added #599 to clean up `active_users`.

PR #631, opened 2022-12-24
update js packages

rawxfish:
I think (?) this updates all the js packages except react-table, which had a patch that seemed somewhat cumbersome to update.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,395 +2,440 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


rgossiaux (approved):
Looks like this may have broken the build though

rawxfish:
Orz

rawxfish:
The build is not broken if I also don't upgrade `webpack-cli`. Why this is the case is beyond my abilities x.x

Issue #616, opened 2022-08-08
transfer ownership now requires consent


asdfryan:
```ERROR: Consent is required to transfer ownership of a file to another user.```

Google Cloud has made it so that when transferring file ownership between users who are not part of the same organization, the transfer requires the new owner to accept the invitation.

Since this requires the new owner to accept the invitation, the transfer cannot be done programmatically via API, so this method fails for users who don't belong to the same organization.

We need the ownership transfer to make scripts work.

rawxfish:
A work around is to pre-make ~200 sheets before the hunt and make sure ownership is properly transferred. This just adds an additional click per sheet to MITMH 2022, when we had to pre-make the sheets to get the formulas to work.

PR #632, opened 2022-12-25
fix iss #616: pending owner

rawxfish:
Asks new owner for permission before transferring. 

It seems that this does not send an "Invitation to own" email as it would if the transfer was done through the normal Google drive UI.  Instead, you can see all the sheets with pending transfers by searching `pendingowner:me` or https://drive.google.com/drive/u/1/search?q=pendingowner:me. Then, select all, click "Share" and click "Accept Ownership?"

image

There's no longer need to do this mid-hunt, since the new named functions can be run even if it is owned by service account. We can do this after the hunt to make sure the sheets are owned by a normal account (not a huge deal if we don't, just prevents the sheets from being lost into the abyss if we ever get rid of the service account).

### DIFF
--- a/google_api_lib/tasks.py
+++ b/google_api_lib/tasks.py
@@ -78,8 +78,7 @@ def transfer_ownership(self, file, template_file_id) -> None:
     self.drive_service().permissions().update(
         fileId=file["id"],
         permissionId=permission["id"],
-        body={"role": "owner"},
-        transferOwnership=True,
+        body={"role": "writer", "pendingOwner": "true"},
     ).execute()
 
 


google_api_lib/tasks.py
```diff
@@ -78,8 +78,7 @@ def transfer_ownership(self, file, template_file_id) -> None:
     self.drive_service().permissions().update(
         fileId=file["id"],
         permissionId=permission["id"],
-        body={"role": "owner"},
-        transferOwnership=True,
+        body={"role": "writer", "pendingOwner": "true"},
```
erwa, line 81: any reason for changing role from "owner" to "writer"?

erwa (approved):

google_api_lib/tasks.py
rawxfish: "owner" is no longer allowed, have to do pendingOwner instead, but role is still a required field, so =/

PR #634, opened 2022-12-25
add feeders on meta metas

rawxfish:
image


google_api_lib/tasks.py
```diff
@@ -222,8 +222,62 @@ def __clear_sheet(sheets_service, http, spreadsheet_id, sheet_id) -> None:
     ).execute(http=http)
 
 
+def _get_feeder_table(feeders) -> int:
```
erwa, line 225: nit: maybe rename this to _build_feeder_table since it's just constructing an object based on the input feeders but not actually "fetching" anything (which I think of when I see "get").

Alternatively, perhaps this method could take as input the Meta puzzle object itself, and we could do the puzzle.feeders.all() call inside here.

erwa (approved):
Nice!

PR #639, opened 2022-12-27
add linting, resolve certain lint violations

pcvera:
I turned off the rules we're violating a lot to keep the PR diff small. We can re-enable them if we're feeling particularly strongly about them.

We can add this to CI in a future update.


pcvera:
looking into why the pre-commit prettier execution conflicts with the local version

pcvera:
figured it out, pre-commit was stuck to an older version of prettier, I updated the version and now the local version and the pre-commit version are consistent

erwa (approved):
Looks good to me. Yay linting!

PR #642, opened 2022-12-27
move default tag setup to BE

rawxfish:


rgossiaux (changes requested):
Nice!

I think this changes the order of the tags in the modal, right? So now they'll be sorted in the order they were created in instead of before where it was default tags first, then custom tags sorted by color. Not sure if that's intentional or not. Probably doesn't really matter for custom tags but I think it also affects the meta tags.

The other maybe more serious thing that I wanted to check is IIRC we delete tags when we remove them from the last puzzle they were on, is that right? In that case tagging and untagging a puzzle with a default tag might kill the default tag.

Maybe we should add another field to record whether it's a default tag or something? That would let us handle the ordering as well as the deletion thing.

rawxfish:
Yeah the order will be a little different. The meta tags are separated out already, so just the custom ones after that will be sorted by creation time, which shouldn't be so bad
image
Good point about the tag deletion, added a `is_default` field as you suggested

rgossiaux (approved):
Ah cool, forgot that we dealt with the meta tags separately. lgtm although maybe if you're feeling tryhard we should add a unit test or two for this

PR #643, opened 2022-12-28
add hunt settings page

rawxfish:
image

Accessible to staff users from hunt index page. Also redirected here after hunt creation (which I also gated to only staff). We aren't tracking who created a hunt currently, so staff seemed like the best proxy.


rgossiaux (approved):
crushing it

PR #647, opened 2022-12-28
discord part 5: Delete channels and rename

asdfryan:


rawxfish (approved):

chat/tests.py
```diff
@@ -194,7 +194,12 @@ def __init__(self, django_settings):
                     func("channel", "name")
                 elif f == "categorize_channel":
                     func("guild-id", "channel-name-or-id", "category-name")
-                elif f == "get_text_channel_participants":
+                elif f in [
+                    "get_text_channel_participants",
+                    "delete_channel",
+                    "delete_audio_channel",
+                    "delete_text_channel",
```
erwa, line 201: This test looks like it verifies that all of these methods raise NotImplementedError (this code is under `with self.assertRaises(NotImplementedError)`). But it also looks like in this PR, you've implemented `delete_text_channel`. What's going on?

erwa: Oh nvm, I see you are testing PartiallyImplementedChatService which doesn't implement any of these methods.

PR #649, opened 2022-12-29
add location tags

### LINKED ISSUE #621: Add location tags?
Maybe like a 

- Classroom A
- Classroom B
- Remote

Bonus: automatically untags itself if there is no activity in the sheets for a certain period of time.

rawxfish:
image

closes #621


api/tests.py
```diff
@@ -314,6 +326,41 @@ def test_cannot_change_meta_tag_color(self):
         meta_tag = puzzle.tags.get(name=meta.name)
         self.assertEqual(meta_tag.color, PuzzleTag.BLACK)
 
+    def test_empty_custom_tags_deletion(self):
```
rawxfish, line 329: moved these two tests from puzzles/test.py, since they're testing code in api/views.py and not stuff in puzzles/

rgossiaux (changes requested):
Nice

Just noticed it's kind of weird to have two different fields that indicate whether a tag is a location (color == teal and is_location) and are using both as ground truth in different places, I think we should a db constraint to make sure these stay consistent with each other. Ditto for the other boolean fields on the tags, actually. They're pretty easy to add in django: https://docs.djangoproject.com/en/4.1/ref/models/constraints/#checkconstraint

rgossiaux (approved):

puzzles/signals/handlers.py
```diff
@@ -2,7 +2,7 @@
 from django.db.models.signals import pre_save, post_save, pre_delete, m2m_changed
 from django.dispatch import receiver
 from puzzles.models import Puzzle
-from puzzles.puzzle_tag import PuzzleTag, PuzzleTagColor
+from puzzles.puzzle_tag import PuzzleTag, PuzzleTagColor, META_COLOR
```
erwa, line 5: nit: it looks like the PuzzleTagColor import is no longer used after directly importing META_COLOR

rawxfish: removed it as a part of #655

PR #652, opened 2022-12-30
use better chevrons for the expansion toggle

pcvera:
image

Before, for reference
image

fixes https://github.com/cardinalitypuzzles/cardboard/issues/597


rgossiaux (approved):
lol your system renders the old ones terribly

PR #654, opened 2023-01-02
use sheet activity to track active users

rawxfish:
Sheet activity is queried as a celery task every N minutes from the [Google Drive Activity API](https://developers.google.com/drive/activity/v2/reference/rest/v2/activity/query), which seemed to return pretty granular edit activities (maybe too granular). N can be controlled via the admin panel. I tried to be somewhat careful to not have this bring down the db with spamming queries, but I didn't test with >50 users and >100 puzzles. In case this feature starts wreaking havoc, the job can also be disabled through admin panels:

image

I also currently have it in a separate job queue to avoid having this feature from blocking discord and sheet creation. And also it was an easy way to make sure only one copy of the job ran at once since I was too lazy to figure out how redis locks worked lol.

The API has new field `recent_editors` that's just a list of names. Recent editors is the list of people who have has an edit activity in the past M time, which is configurable in hunt settings.


Procfile
```diff
@@ -1,4 +1,5 @@
 release: python manage.py migrate --noinput
 web: gunicorn cardboard.wsgi --log-file -
 worker: celery -A cardboard worker -l INFO --without-heartbeat --without-gossip --without-mingle --concurrency 2
+worker: celery -A cardboard worker -l INFO --without-heartbeat --without-gossip --without-mingle --concurrency 1 -Q activity_updates_queue --beat
```
erwa, line 4: Does this worker for the Google Drive activity run on the same dyno as the other worker? Just curious if this requires another dyno. For saving money, it'd be nice if it could run in the same dyno as the other worker.

erwa (approved):
Really nice! Excited to see the active users rendered in the UI -- will be really cool and useful.

Procfile
rawxfish: Different dyno, I think it's better for it to be separate, at least this year, since this is a feature that might block more important things that the other celery workers are doing, like sheets creation.

rgossiaux: i think our celery setup is pretty wasteful since all of the dynos should be idle almost all the time, but we can probably work on that separately

google_api_lib/tasks.py
```diff
@@ -426,3 +437,172 @@ def update_meta_and_metameta_sheets_delayed(meta):
...
+                    last_edit_time=timestamp,
+                )
+                for ((user_pk, puzzle_pk), timestamp) in latest.items()
+                if (user_pk, puzzle_pk) not in old_keys
+            ],
+        )
+
+        updates = []
+        for activity in old_activities:
+            activity.last_edit_time = latest[(activity.user_id, activity.puzzle_id)]
```
rgossiaux, line 606: is it possible for this user/puzzle pair to not exist in `latest`? won't this cause an issue?

rawxfish: `old_activities` was formed from the keys of latest via the filtering in lines 574-585, so it should always exist I think? Admittedly, the filtering feels sorta hacky, but I have no idea how to make it better.

PR #656, opened 2023-01-03
add silk for profiling when DEBUG is true

rawxfish:


rgossiaux (approved):

cardboard/settings.py
```diff
@@ -82,6 +82,11 @@
     "social_django.middleware.SocialAuthExceptionMiddleware",
 ]
 
+
+if DEBUG:
+    INSTALLED_APPS += ["silk"]
+    MIDDLEWARE = ["silk.middleware.SilkyMiddleware"] + MIDDLEWARE
```
erwa, line 88: Is it important that the silk middleware comes first or could we use MIDDLEWARE.append("silk.middleware.SilkyMiddleware") (or +=)?

(Also got curious about += vs. append(): https://stackoverflow.com/questions/725782/in-python-what-is-the-difference-between-append-and - apparently append() is faster)

rawxfish: Yes the middleware placement is more sensitive: https://github.com/jazzband/django-silk

PR #658, opened 2023-01-04
improve performance of default puzzle tag population

rawxfish:
Old code was doing a separate db query for each `get_or_create` on each default tag.

### DIFF
--- a/puzzles/puzzle_tag.py
+++ b/puzzles/puzzle_tag.py
@@ -105,15 +105,31 @@ class PuzzleTag(models.Model):
 
     @staticmethod
     def create_default_tags(hunt):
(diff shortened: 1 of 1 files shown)


erwa (approved):
Sure, though it looks like create_default_tags() is only called when a new hunt is created and then never again so not sure how much performance matters?

rawxfish:
It's also called when you edit the hunt (it's sorta a hacky way to populate or de-populate the default tags)

erwa:
Gotcha. Thanks for optimizing this!

rgossiaux (approved):

erwa:
oh, btw, did you find this performance issue after enabling silk in #656?

rawxfish:
yes

PR #659, opened 2023-01-04
Add indicator for recent sheet editors

pcvera:
image

image

Hover
image

image


hunts/src/puzzle-table.js
```diff
@@ -27,6 +28,12 @@ const TABLE_COLUMNS = [
     Cell: NameCell,
     className: "col-4",
   },
+  {
+    Header: "",
+    accessor: (row) => row.recent_editors,
```
erwa, line 33: Is this for sorting by this column? Will this sort from most to least recent editors?

Also, will the search box allow you to filter puzzles by who is working on them? (say, I really want to puzzle with Joe)

hunts/src/UsersCell.js
```diff
@@ -0,0 +1,55 @@
...
+        placement="right"
+        overlay={recentEditorsPopover}
+      >
+        <div
+          className={getColor(recentEditors.length)}
+          style={{
+            whiteSpace: "nowrap",
+          }}
+        >
+          <IconUsers size={16} />
```
erwa, line 47: Do the icons look ok on the Light background too?

erwa (approved):
Looks great! (though I'm hardly qualified to review JS/React code)

erwa:
Also, can you post a screenshot of what the hoverover pop-up listing the recent editors look like?

rgossiaux:
Cool, thanks for doing this. Might be easier to scan if we limit to fewer colors I think, the distinction between eg 3 and 4 people on the sheet is probably not super relevant. Maybe three ranges that roughly approximate "not full", "full" and "overfull"? Maybe like 1-4, 5-8, and 9+ solvers or something (just pulled these out of my ass)? I guess the exact thresholds will depend on team size and # of open puzzles and could maybe be guesstimated dynamically but we could hardcode something for now.

asdfryan:
It'd be nice to have an `updated 3 mins ago` or something on hover

PR #660, opened 2023-01-05
few improvements to active users

rawxfish:
A few changes to make active users backwards compatible or more robust.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 release: python manage.py migrate --noinput
 web: gunicorn cardboard.wsgi --log-file -
 worker: celery -A cardboard worker -l INFO --without-heartbeat --without-gossip --without-mingle --concurrency 2
-worker: celery -A cardboard worker -l INFO --without-heartbeat --without-gossip --without-mingle --concurrency 1 -Q activity_updates_queue --beat
+activity_monitor: celery -A cardboard worker -l INFO --without-heartbeat --without-gossip --without-mingle --concurrency 1 -Q activity_updates_queue --beat
 bot: python manage.py rundiscordbot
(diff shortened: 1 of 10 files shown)


Procfile
```diff
@@ -1,5 +1,4 @@
 release: python manage.py migrate --noinput
 web: gunicorn cardboard.wsgi --log-file -
-worker: celery -A cardboard worker -l INFO --without-heartbeat --without-gossip --without-mingle --concurrency 2
```
erwa, line 3: Nice, glad you were able to combine the workers

scripts/add_service_account_contacts.py
```diff
@@ -0,0 +1,35 @@
+import datetime
+
+import googleapiclient
+from django.conf import settings
+from google.oauth2 import service_account
+
+from accounts.models import Puzzler
+
+# For adding all registered user into service account's contact list
```
erwa, line 9: user -> users

google_api_lib/tasks.py
```diff
@@ -461,21 +462,48 @@ def extract_id_from_drive_item_name(drive_item_name) -> Optional[str]:
...
     UserModel = get_user_model()
     try:
         return UserModel.objects.get(social_auth__uid=uid)
     except UserModel.DoesNotExist:
+        pass
+
+    # Fall back to People API
+    # Note that this API depends on the email being added as service account's contact list
+    # It also has a tight request limit
+    time.sleep(2)
```
erwa, line 475: Why do we need to sleep here?

rawxfish: So this branch of the code should only run for backward compatibility with previous year's data. When that happens, we actually call the People API in quick succession and it's easy to hit the API's 90 request / minute quota, probably could do with just 1 second though.

asdfryan (approved):

google_api_lib/tasks.py
erwa: Oh gotcha. Could you mention that the sleep is to prevent exceeding Google API limits?

erwa: I guess you mentioned the tight request limit, nvm

PR #664, opened 2023-01-06
Optional discord

asdfryan:
https://www.loom.com/share/9b7122d9eb3342be99f234c8e261f95d


rawxfish (approved):

erwa:
The default is no Discord channel creation, right? Just wondering if we want to support making channel creation on the default. Did people feel most puzzles' Discord channels weren't used, so it wasn't worthwhile to create Discord channels by default?

rawxfish:
We don't expect them to be used that much this year since a majority of the team will be onsite and will just cause extra discord clutter and notifs.

PR #665, opened 2023-01-07
add more retries, increase time limit for sheet creation

rawxfish:


erwa (approved):
Did you hit some errors while testing something out?

rawxfish:
@asdfryan did. I also recall needing to manually create some sheets during the past mystery hunts.

PR #666, opened 2023-01-07
Adds settings variables to control app branding.

rohit-rao:
(Continuing to upstream my local diffs.)

When I set up a site for my team, it felt wrong to keep the "Cardinality" branding. If you agree, here's an attempt to make the branding configurable. I've kept the defaults unchanged, so there should be no visible changes unless you set the appropriate env variables.

### DIFF
--- a/cardboard/context_processors.py
+++ b/cardboard/context_processors.py
@@ -9,5 +9,10 @@ def google_auth(request):
     }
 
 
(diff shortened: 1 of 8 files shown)


rawxfish (approved):

erwa:
Nice! Thanks for updating the docs too.

PR #669, opened 2023-01-11
whoops

rawxfish:


asdfryan (approved):

chat/models.py
```diff
@@ -176,7 +176,9 @@ def delete_channels(self, check_if_used=False):
                 participants = service.get_text_channel_participants(
                     self.text_channel_id
                 )
-                should_delete_text_channel = participants and len(participants) == 0
+                should_delete_text_channel = (
+                    participants is not None and len(participants) == 0
```
erwa, line 180: Should this be `participants is None or len(participants) == 0`? If participants is None, we _should_ delete the text channel, right?

rawxfish: If participants is None, then the API bugged out, so we should be cautious with deleting.

erwa: What's the difference between the new and old versions then? In the old version (`participants and len(participants) == 0`), if participants was None, `should_delete_text_channel` would still evaluate to `False`, right?

rawxfish: It is different for `participants = []`

erwa: Ah, I see. I think it'd be helpful to add a comment about this subtlety

PR #686, opened 2023-02-17
debug default to false / fix precommit

asdfryan:


.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/pycqa/isort
-    rev: 5.11.2
+    rev: 5.12.0
```
asdfryan, line 20: https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co

rawxfish (approved):

.pre-commit-config.yaml
akirabaruah: Nice. Thanks!

PR #710, opened 2023-12-13
Minor logging changes from Puzzle Boat 10 (2023)

akirabaruah:


cardboard/settings.py
```diff
@@ -311,6 +311,10 @@ class TaskPriority(Enum):
         },
     },
     "loggers": {
+        "": {
+            "handlers": ["console"],
+            "level": os.getenv("DJANGO_LOG_LEVEL", "INFO"),
```
rgossiaux, line 316: I think this env variable is supposed to be specific to the logging messages that django itself generates so I don't think we should reuse it here as this is for the logging messages we created ourselves. IMO better to either hardcode it, condition it on the DEBUG setting, or create a separate env variable

cardboard/settings.py
```diff
@@ -311,6 +311,10 @@ class TaskPriority(Enum):
         },
     },
     "loggers": {
+        "": {
```
rgossiaux, line 314: it looks like you can alternatively pull this out to the top level in a `root` entry a la https://docs.djangoproject.com/en/5.0/topics/logging/#id3 

don't think there's any functional difference between the two

akirabaruah: Yep. Done.

cardboard/settings.py
akirabaruah: Sounds good. I just changed the variable to `DJANGO_APP_LOG_LEVEL` for flexibility in the future.

rgossiaux (approved):
sweet, thanks!

PR #711, opened 2023-12-21
Fix docker-compose setup

akirabaruah:
The `docker-compose` workflow seems broken without these updates.


cardboard/settings.py
```diff
@@ -155,6 +155,10 @@
 ]
 STATICFILES_STORAGE = "whitenoise.django.CompressedManifestStaticFilesStorage"
 
+# TODO: Fix the missing manifest issues. This is a workaround.
+# https://github.com/cardinalitypuzzles/cardboard/issues/712
+WHITENOISE_MANIFEST_STRICT = False
```
rgossiaux, line 160: I think we should probably just leave this alone, the `collectstatic` change should almost certainly fix this

akirabaruah: Unfortunately, if I remove this, I still see the `ValueError: Missing staticfiles manifest entry for 'favicon.ico'` message at run time. So I think there's a couple layers to this problem:

1. Actually generating static files and associated paths, which was covered by `collectstatic`.
2. Configuring the manifest correctly. I'm not sure how to debug that bit, but turning off `WHITENOISE_MANIFEST_STRICT` does seem to workaround it until we properly fix it?

akirabaruah: To unblock devs, I think we should submit this workaround, but then try to fix the manifest issue ASAP. I'm happy to keep digging into it for https://github.com/cardinalitypuzzles/cardboard/issues/712 (can add it to MITMH 2024 milestone and assign myself), unless you have other ideas.

rgossiaux: well I think the problem is that you need to run `collectstatic` on your machine. If you're saying that the docker change doesn't actually fix that for you, then we should undo the docker change, although I'm surprised that isn't enough--did you restart the container? 

In any case, when the container is running, you can also run `docker-compose exec web python manage.py collectstatic` to run collectstatic manually

akirabaruah: > If you're saying that the docker change doesn't actually fix that for you, then we should undo the docker change, although I'm surprised that isn't enough--did you restart the container?

I tested each change in isolation (making sure to stop, prune, and restart containers each time) and found that we need both changes to avoid the issue.

But it looks like @kcaze found the proper fix for the 2nd issue in https://github.com/cardinalitypuzzles/cardboard/pull/715, so we can close this PR and merge that one.

akirabaruah:
See https://github.com/cardinalitypuzzles/cardboard/pull/715

PR #715, opened 2023-12-21
Staticfiles fix

kcaze:
It looks like there were 2 main issues with static files:
1. The staticfiles configs we set were getting clobbered by `django_heroku.settings()`. Setting `staticfiles=False` means we're using our own settings.
2. The staticfile config we did have wasn't correct. It was missing `STATIC_ROOT` and the backend storage config was also incorrect. See https://whitenoise.readthedocs.io/en/latest/django.html for details.

I also updated docker-entrypoint.sh to use the `--no-input` flag when running `collectstatic`.


cardboard/settings.py
```diff
@@ -146,14 +146,19 @@
...
 STATICFILES_DIRS = [
     os.path.join(BASE_DIR, "cardboard/static"),
     os.path.join(BASE_DIR, "hunts/static"),
 ]
-STATICFILES_STORAGE = "whitenoise.django.CompressedManifestStaticFilesStorage"
+STORAGE = {
+    "staticfiles": {
+        "BACKEND": "whitenoise.storage.CompressedManifestStaticFilesStorage",
+    }
+}
```
akirabaruah, line 161: Is this change required? Or is it basically the "new way" to do things? (in which case, seems reasonable to switch)

cardboard/settings.py
```diff
@@ -172,7 +177,7 @@
 # Configure Django App for Heroku.
 import django_heroku
 
-django_heroku.settings(locals(), test_runner=False)
+django_heroku.settings(locals(), test_runner=False, staticfiles=False)
```
akirabaruah, line 180: Should we add a comment above this line pointing out that `staticfiles=False` is to avoid clobbering previous settings?

akirabaruah (approved):
Dope 😎 Thanks for the fix! Had some minor suggestions but otherwise LGTM

cardboard/settings.py
```diff
@@ -146,14 +146,19 @@
...
+# https://whitenoise.readthedocs.io/en/latest/django.html
 
+STATIC_ROOT = os.path.join(BASE_DIR, "staticfiles")
 STATIC_URL = "/static/"
 STATICFILES_DIRS = [
     os.path.join(BASE_DIR, "cardboard/static"),
     os.path.join(BASE_DIR, "hunts/static"),
 ]
-STATICFILES_STORAGE = "whitenoise.django.CompressedManifestStaticFilesStorage"
+STORAGE = {
```
rgossiaux, line 157: This is a typo, this setting is named `STORAGES` https://docs.djangoproject.com/en/5.0/ref/settings/#storages

But it was also added in Django 4.2. In the `pyproject.toml` we only have 4.1 specified and the package version in the `poetry.lock` is also 4.1, so you should use the 4.1 settings (we should also upgrade Django to the latest version but probably in a separate PR)

cardboard/settings.py
kcaze: I tested with this syntax vs the other and it seems necessary. According to the docs, this is needed for Django 4.2+

cardboard/settings.py
rgossiaux: Where did this change come from?

rgossiaux (changes requested):

cardboard/settings.py
kcaze: Oh oops, you're right. I'll delete my `staticfiles` directory and test from scratch again with this fixed.

rgossiaux:
I think our settings are working ok and `django_heroku` isn't doing anything too nefarious to them; you can see what changes it makes here: https://github.com/heroku/django-heroku/blob/master/django_heroku/core.py We have a few redundant lines in this file but nothing in conflict.

With that said we should probably just rip out that module wholesale and inline everything. That's what the creator recommends; looks like it's been deprecated for a couple years: https://github.com/heroku/django-heroku/issues/56

cardboard/settings.py
akirabaruah: Sounds good. FWIW, I think Django is still 4.1 in the poetry.lock file.

akirabaruah:
Now that we have the webpack fix in https://github.com/cardinalitypuzzles/cardboard/pull/716, we can prob close this PR and followup with a separate `django-heroku` cleanup one?

PR #716, opened 2023-12-22
fix dev environment

rgossiaux:
`collectstatic` has been kind of broken in dev because the main.js file generated by webpack references some source map files in node_modules, and Django doesn't like the references to files that it doesn't know about.

This webpack plugin extracts these source map references, which fixes the collectstatic issue and probably also fixes source maps for 3rd party libraries if they were broken.


kcaze:
Nice, this looks good to me. I had noticed the sourcemap issue occurring at one point as well but didn't know how to address it.

Shall we remove the usage the `django_heroku` in a separate PR and then also pin the django version?

akirabaruah (approved):
Nice! LGTM

akirabaruah:
> Nice, this looks good to me. I had noticed the sourcemap issue occurring at one point as well but didn't know how to address it.
> 
> Shall we remove the usage the `django_heroku` in a separate PR and then also pin the django version?

Yeah I think it'd make sense to clean up the django_heroku stuff in a separate PR since it's somewhat orthogonal to this change.

PR #717, opened 2023-12-22
Add DEFAULT_AUTO_FIELD setting

rgossiaux:
This was added in Django 3.2 and was creating a lot of warnings that were clutting the Docker output.


akirabaruah (approved):
Ah thanks for cleaning this up. I'll go ahead and merge since it's pretty straightforward.

PR #718, opened 2023-12-22
Update virtualenv version

akirabaruah:
Without this change, `docker-compose up -d --build` tries to downgrade `virtualenv` in the based Docker image as seen by this log:

```
• Downgrading virtualenv (20.25.0 -> 20.17.1)
```

This causes the build to fail due to nonexistent virtualenv submodules.


rgossiaux (approved):
Interesting, works fine for me but seems harmless enough regardless

Issue #667, opened 2023-01-10
Move multiple answers option to inside the edit answers modal


rgossiaux:
Ever since we added this feature it's irritated me that the primary UI is cluttered with so many buttons that are only relevant in the rare hunt that has multiple answers per puzzle:

<img width="888" alt="image" src="https://user-images.githubusercontent.com/3475509/211476060-5697bd4f-ad21-407c-b0f7-21950a394040.png">

Instead, we could just remove these + buttons and change the edit answer modal to also let you add/edit multiple answers.

rgossiaux:
(related to #280 )

Issue #280, opened 2021-01-02
De-clutter React UI


asdfryan:
Context: https://github.com/cardinalitypuzzles/smallboard/pull/275#discussion_r549605998

![smallboard-ui](https://user-images.githubusercontent.com/1498449/103448160-5b6b9480-4c4a-11eb-94fa-266dc8f28b33.png)

We've got a lot of buttons on our UI for editing/adding/deleting puzzles and answers. We can de-clutter this by removing these buttons and exposing these functionalities not on the table itself, but on the modal dialog instead.

erwa:
Note the screenshot above is out-of-date; the Tags and Meta columns have been combined into one now. There's also a "Chat" column that has a link to the Discord text channel for the puzzle.

rgossiaux:
We've streamlined this enough that I'm just going to close this issue

PR #720, opened 2023-12-25
Display puzzle name and answer editing buttons on hover.

kcaze:
Addresses #667 and #280 by making the editing buttons appear only on hover. This significantly reduces the UI clutter.

https://github.com/cardinalitypuzzles/cardboard/assets/342514/bb3419f4-89c1-4104-87fb-470d634aadc7


rgossiaux (approved):
Nice! Looks a lot cleaner to me.

I think ideally this would be implemented in CSS but we don't actually have any CSS files..... but maybe you could put a TODO?

kcaze:
I think for css in general, it might be good to use tailwind.

Issue #704, opened 2023-12-11
Display long answers better


rawxfish:
"max-width on the answer length"

kcaze:
Probably a max-width property + `overflow-wrap: anywhere`.

rgossiaux:
@kcaze fixed in #721!

PR #721, opened 2023-12-29
Max answer width

kcaze:
Address #704 by setting a max-width of 400px for answers and wrapping them if they exceed the max width.

Before:
before

After:
after


rgossiaux (approved):
we DO have a CSS file.... lol. I thought so but for some reason didn't find it the other day. anyway lg

Issue #672, opened 2023-01-13
Meta tags don't appear right away after adding a new meta


rgossiaux:
We noticed that when you add a new meta in cardboard, the tag corresponding to it doesn't appear right away if you create a new puzzle and try to assign it to the new meta.

rgossiaux:
They actually won't appear at all until you refresh. It's a regression from https://github.com/cardinalitypuzzles/cardboard/pull/642/files#diff-195bf4ead66185014cd09d75a1621afd4bb647781a8bca3af807ad1a3c141403

We used to derive the list of tags from the tags on each individual puzzle. Now we're getting the list of tags from the Hunt object, but unlike the puzzles we don't refresh the hunt, so the tags never update after you reload the page.

rgossiaux:
This was fixed by #755

rgossiaux:
This was reported as not fixed after all

akirabaruah:
I was unable to reproduce this locally at current HEAD. Creating sample Meta puzzles immediately populates available tags for puzzles without refresh.

I did notice that the Meta list didn't populate correctly when subsequently adding feeders, but that seems like a different issue. Refreshing the page resolved that for me.

PR #722, opened 2023-12-29
Refresh tags upon adding a puzzle

kcaze:
Address #672.

This re-fetches the hunt slice after adding a puzzle, so the new meta tag appears.

### DIFF
--- a/hunts/src/AddPuzzleModal.js
+++ b/hunts/src/AddPuzzleModal.js
@@ -5,6 +5,7 @@ import Button from "react-bootstrap/Button";
 import Form from "react-bootstrap/Form";
 import { addPuzzle } from "./puzzlesSlice";
 import { hideModal } from "./modalSlice";
+import { fetchHunt } from "./huntSlice";
 
 function AddPuzzleModal({ huntId }) {
   const [name, setName] = React.useState("");
@@ -23,6 +24,8 @@ function AddPuzzleModal({ huntId }) {
         create_channels: createChannels,
       })
     ).finally(() => {
+      // Used to update hunt tags.
+      dispatch(fetchHunt(huntId));
       dispatch(hideModal());
     });
     return false;
--- a/hunts/src/EditPuzzleModal.js
+++ b/hunts/src/EditPuzzleModal.js
@@ -3,7 +3,7 @@ import { useDispatch } from "react-redux";
 import Modal from "react-bootstrap/Modal";
 import Button from "react-bootstrap/Button";
 import Form from "react-bootstrap/Form";
-import { updatePuzzle } from "./puzzlesSlice";
+import { updatePuzzle, fetchPuzzles } from "./puzzlesSlice";
 import { hideModal } from "./modalSlice";
 import { fetchHunt } from "./huntSlice";
 
@@ -29,6 +29,8 @@ function EditPuzzleModal({ huntId, puzzleId, name, url, isMeta, hasChannels }) {
     ).finally(() => {
       // Used to update hunt tags.
       dispatch(fetchHunt(huntId));
+      dispatch(fetchPuzzles(huntId));
+
       dispatch(hideModal());
     });
     return false;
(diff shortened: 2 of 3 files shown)


rgossiaux:
Does this fully fix the issue? I haven't tested it but from my understanding the tags are also missing when other people add metas (or if they create new tags in some other way)

I think we might want to just refresh the Hunt object periodically, like every 60s or so, similar to what we're doing with refreshing the puzzles. I think this change you're making to refresh it immediately after adding a puzzle is also good for the case when you add a meta and then immediately try to add a feeder puzzle to that meta, though we could check if it's a meta first as an optimization. For correctness we would also want to do an immediate refresh if you edit the puzzle to toggle whether it's a meta, I think.

kcaze:
Yeah that makes sense. I'll add a periodic refresh and also a refresh when you edit tags.

kcaze:
Tested editing a puzzle to change it to a meta and then change it to not a meta.

I opened two windows to test the periodic refresh for tags.

hunts/src/EditPuzzleModal.js
```diff
@@ -26,6 +27,10 @@ function EditPuzzleModal({ huntId, puzzleId, name, url, isMeta, hasChannels }) {
         },
       })
     ).finally(() => {
+      // Used to update hunt tags.
+      dispatch(fetchHunt(huntId));
+      dispatch(fetchPuzzles(huntId));
```
rgossiaux, line 32: Don't think this is necessary, is it?

hunts/src/AddPuzzleModal.js
```diff
@@ -23,6 +24,8 @@ function AddPuzzleModal({ huntId }) {
         create_channels: createChannels,
       })
     ).finally(() => {
+      // Used to update hunt tags.
+      dispatch(fetchHunt(huntId));
```
rgossiaux, line 28: I think only if `isMeta` is true do we need to do this, right?

hunts/src/EditPuzzleModal.js
```diff
@@ -26,6 +27,10 @@ function EditPuzzleModal({ huntId, puzzleId, name, url, isMeta, hasChannels }) {
         },
       })
     ).finally(() => {
+      // Used to update hunt tags.
+      dispatch(fetchHunt(huntId));
```
rgossiaux, line 31: Same as above, we only need to do this if `isMeta` is different from `newIsMeta` right?

hunts/src/HuntViewMain.js
```diff
@@ -49,16 +49,17 @@ export const HuntViewMain = (props) => {
 
   const dispatch = useDispatch();
 
-  const updatePuzzleData = () => {
+  const updatePuzzleAndHuntData = () => {
     dispatch(fetchPuzzles(props.huntId));
+    dispatch(fetchHunt(props.huntId));
```
rgossiaux, line 54: This might add a lot of load if we do it every 10 seconds

rgossiaux (commented):
Actually stepping back for a second (sorry for the churn), does this change make sense to begin with? I kind of feel like what we should be doing is just storing the default tags on the `Hunt` object (which is fetched once when you load the page and is not expected to change), and then reading the list of available tags from the list of puzzles kind of like we did before: 

https://github.com/cardinalitypuzzles/cardboard/pull/642/files#diff-2d036991b0fa7e0ecb01d9f9c592d66f59257862e92de3635f04f1555b19a000L231

But that logic can be simplified a lot now to just select all the tags from the `Puzzle` objects, combine with all the tags on the `Hunt` object, dedupe and sort by tag id.

Does that make sense?

kcaze:
Yeah that makes sense, I don't have a good sense of how much additional load fetching the hunt object is (since we already fetch puzzles every 10 seconds). From a perf standpoint, I guess ideally we have a websocket connection with which the server pushes updates?

It makes sense to me to merge the default tags from the hunt object + custom tags from the puzzles instead. I'll undo these changes and get tags that way.

kcaze:
Closing in favor of #755.

PR #723, opened 2023-12-29
Add freebie tag and tool tip for forward solve, backsolve, and freebie

maximized:


maximized:
Most of the changes to the html file were caused by me running 
```
prettier hunts/templates/stats.html --write --config .prettierrc 
```

hunts/models.py
```diff
@@ -47,6 +47,14 @@ def get_object_or_404(user=None, **kwargs):
     def get_num_solved(self):
         return self.puzzles.filter(status=Puzzle.SOLVED).count()
 
+    def get_num_backsolved(self):
+        all_solved = self.puzzles.filter(status=Puzzle.SOLVED)
+        return sum(1 for puzzle in all_solved if puzzle.is_backsolved())
```
rgossiaux, line 52: I believe this is going to do a fresh db query per puzzle in order to compute is_backsolved. I think to do one combined query you want to do something like `self.puzzles.filter(status=Puzzle.SOLVED, tags__name=PuzzleTag.BACKSOLVED).count()` (I haven't tested this though). ditto below

rgossiaux (changes requested):
Nice, thanks!

Just a perf comment (I haven't tested your code so I might be wrong, though)

hunts/models.py
maximized: This worked

maximized (commented):
Addressed comments

rgossiaux (approved):
Ty, lgtm!

PR #724, opened 2024-01-01
Chrome Extension

maximized:


maximized:
https://watch.screencastify.com/v/2XDnV1BxjyDmMmOPf3N6 - Current status

Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN pip install poetry
 RUN poetry config virtualenvs.create false
 COPY pyproject.toml poetry.lock ./
 RUN poetry install
+RUN pip install django-cors-headers
```
rgossiaux, line 23: You should install `django-cors-headers` via `poetry add` and remove this from the Dockerfile

maximized: Yeah, I had some problems with that because certain versions of things didn't match my local versions so I did this for now. I'll fix before requesting a review.

rgossiaux: Ok cool, sounds good. It sounds like you might need to be inside `poetry shell`

maximized:
Waiting for the prod version to be up so I can test that this works in prod too

maximized:
I need the CSRF_TRUSTED_ORIGINS change in settings.py to be checked in to test it with prod.

cardboard/settings.py
```diff
@@ -43,6 +43,9 @@
     "cardinality-cardboard.herokuapp.com",
 ]
 
+# For Cardboard Chrome Extension
+CSRF_TRUSTED_ORIGINS = ["chrome-extension://cahmppnjflkbimomgndbcmbdoafdegbi"]
```
rgossiaux, line 47: Is this guaranteed to be the same for everyone? Or is it different per user?

chrome_extension/cardboard.zip
rgossiaux, line 1: IMO it's better to not check this into the repo, and instead we can release it via github

chrome_extension/cardboard/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
+  "manifest_version": 3,
+  "name": "Cardboard Chrome Extension",
+  "version": "1.0",
+  "icons": {
```
rgossiaux, line 5: If I recall correctly you can just specify one of these if it's the same image (whatever the actual size is) and chrome scales it automatically

chrome_extension/cardboard/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+    "128": "images/cardboard_bird.png"
+  },
+  "action": {
+    "default_popup": "popup.html"
+  },
+  "permissions": [
+    "activeTab",
+	"cookies"
+  ],
+  "host_permissions": [
```
rgossiaux, line 18: I think also 0.0.0.0:8000, at least that's the URL i use with docker

cardboard/settings.py
maximized: This just got published in the Chrome extensions store. The one that everyone will download with have a fixed ID. In development, or if you side load the extension, you will have a different ID. Do you have a problem if we wildcard it instead "chrome-extension://*"?

chrome_extension/cardboard/manifest.json
maximized: I tried one version of this and it continued to show for the Extension itself, but didn't show in the Extensions "store". I can play around with this more in another PR.

chrome_extension/cardboard/manifest.json
maximized: Added

chrome_extension/cardboard.zip
maximized: Removed it

cardboard/settings.py
rgossiaux: Ok got it. I guess as a bandaid we can do the wildcard, it's sort of a security hole but realistically there's not much to attack.

I think the way this "should" work is with token based auth. We could add oauth login to cardboard ([link](https://django-oauth-toolkit.readthedocs.io/en/latest/rest-framework/getting_started.html)) and then use [chrome.identity](https://developer.chrome.com/docs/extensions/reference/api/identity) in the extension to authenticate with cardboard. 

I think it's probably not actually that much work to implement this but we can punt on it for now and track it in a separate issue.

rgossiaux (approved):

PR #732, opened 2024-01-05
Merge tags column into puzzle name; bit of CSS

npinsker:
tweaking the tags UI; in my experience, solvers usually use tags to figure out what puzzles they're interested in solving, so it's made sense for us (in Galactic's software) to put them under titles.

image

### DIFF
--- a/hunts/src/HuntViewMain.js
+++ b/hunts/src/HuntViewMain.js
@@ -73,7 +73,8 @@ export const HuntViewMain = (props) => {
       <SiteHeader />
       <div
         style={{
-          margin: "0 20px",
+          margin: "0px auto",
+          maxWidth: "1600px",
         }}
       >
         <Alert
--- a/hunts/src/NameCell.js
+++ b/hunts/src/NameCell.js
@@ -29,7 +29,7 @@ export default function NameCell({ row, value }) {
   return (
     <div
       style={{
-	      paddingLeft: `${row.depth * 2}rem`,
+        paddingLeft: `${row.depth * 2}rem`,
       }}
     >
       <div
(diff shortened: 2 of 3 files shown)


rgossiaux:
Interesting, thanks for the PR. It might make sense to hide the meta tags in this UI unless the puzzle has >= 2 meta tags, that should make it easier to read and they feel even more redundant to me when you have the meta structure visible right there. I suppose you would also want to display them if the search bar is active since users don't have the context in that case.

akirabaruah:
It seems like there's another trade-off with this change that we introduce slightly more vertical space per puzzle, which can reduce the number of puzzles visible at a time.

Have we considered putting tags to the right of (i.e. inline, but after) each puzzle to reduce vertical space and allow users to scan the list of puzzle names quickly?

Or even simpler: Can we just move the existing Tags column right next to the puzzle Name column?

On the point about redundant meta tagging, I agree that since tags control how meta nesting works, we should keep things simple and accessible to users by showing all tags (including meta tags), for now. If we decide to separate meta puzzle association and generic tagging, that can probably be a separate PR, right?

npinsker:
I'm happy to do either of those changes if you'd prefer it; I was mostly optimizing for appearance.

I'm going to close this PR in favor of https://github.com/cardinalitypuzzles/cardboard/pull/738, which builds on top of this work and mostly addresses the concerns of vertical space. Happy to just have a "tags" column though.

akirabaruah:
Yep, feel free to close this PR in favor of the followup.

Hmm yeah, if it's simple enough to place the tags inline instead of below each puzzle, I'd be interested in seeing what that looks like.

For "scanning", I'd expect we'd want to optimize for vertical scanning (i.e. multiple puzzles), but I agree that shifting your eyes horizontally is also fatiguing. For that reason, I think having a dedicated tags column that's aligned can actually be easier to scan for desired topics. tbh, it's probably a moot point, since we're able to filter puzzles by tags anyway.

PR #735, opened 2024-01-06
add top editors

rawxfish:
Shows the five users with most edits on a sheet, with min of 5 edits.
image

Number of queries is about the same as before:
image

Note that the number of edits shouldn't be taken literally, since Google API batches edits hopefully in a consistent way. The relative values seem okay.. I backfilled last year's hunt and saw that certain puzzles I knew to be big work indeed had larger values for the folks on it


rgossiaux (approved):
nice! looks good to me

PR #736, opened 2024-01-07
Assign a single meta at puzzle creation time

maximized:


maximized:
image

image

image

maximized:
The dropdown for meta options are also subject to https://github.com/cardinalitypuzzles/cardboard/issues/672

hunts/src/AddPuzzleModal.js
```diff
@@ -1,14 +1,17 @@
...
 import Form from "react-bootstrap/Form";
 import { addPuzzle } from "./puzzlesSlice";
 import { hideModal } from "./modalSlice";
+import { selectHuntTags } from "./huntSlice";
 
 function AddPuzzleModal({ huntId }) {
+  const allTags = useSelector(selectHuntTags);
   const [name, setName] = React.useState("");
   const [url, setUrl] = React.useState("");
+  const [assignedMeta, setAssignedMeta] = React.useState("none");
```
rgossiaux, line 14: What's wrong with using a falsy value for this?

chrome_extension/cardboard/manifest.json
```diff
@@ -16,8 +16,8 @@
 	"cookies"
   ],
   "host_permissions": [
-	"0.0.0.0:*",
-	"http://localhost:*/*",
+	"http://*:*/*",
```
rgossiaux, line 19: This feels kind of sketchy, was this just to include 127.0.0.1 or was there more? I guess it's ok as a temporary hack if we need to

hunts/src/AddPuzzleModal.js
```diff
@@ -53,6 +58,25 @@ function AddPuzzleModal({ huntId }) {
               onChange={(e) => setUrl(e.target.value)}
             />
           </Form.Group>
+          <Form.Group controlId="addPuzzle.meta">
+            <Form.Label>Assigned Meta</Form.Label>
+            <Form.Control
+              as="select"
+              value={assignedMeta}
```
akirabaruah, line 65: If you use a falsy value like `null` (@rgossiaux's comment above), you can probably change this to `value={assignedMeta ?? "none")`.

chrome_extension/cardboard/manifest.json
akirabaruah: Also, would this potentially get flagged during the Chrome Extension store review?

api/views.py
```diff
@@ -375,6 +375,16 @@ def create(self, request, **kwargs):
...
 
+            if (
+                "assigned_meta" in request.data
+                and request.data["assigned_meta"]
+                and request.data["assigned_meta"] != "none"
+            ):
+                meta = get_object_or_404(
+                    Puzzle, name=request.data["assigned_meta"], hunt=puzzle.hunt
+                )
+                puzzle.metas.add(meta)
```
akirabaruah, line 386: Should we add a unit test for this? (Can be in a separate PR if we want to expedite this one)

hunts/src/AddPuzzleModal.js
maximized: Nothing. Should I use the falsy value for the none option too?

chrome_extension/cardboard/manifest.json
maximized: @rgossiaux, didn't you say you use 0.0.0.0? 

@akirabaruah I actually already included this in the most recent extension upload and it was accepted. However, maybe this is why it took longer.

https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns has Chrome's suggestions. We can include just `http://localhost/*`, `http://127.0.0.1/*`, and `http://0.0.0.0/*`. Their documentation also suggests:

```
You can also use http://*:*/* to match localhost, IP addresses, and any port.
```

api/views.py
maximized: Sure. I'll add a unit test.

hunts/src/AddPuzzleModal.js
maximized: I'll just change both to "".

chrome_extension/cardboard/manifest.json
akirabaruah: Ah thanks for the clarifications. SGTM.

I guess we want the `*`s because we want to run this extension on any website, i.e. arbitrary puzzle pages?

akirabaruah (approved):
LGTM with pending changes.

chrome_extension/cardboard/manifest.json
maximized: No. I'm not authorizing using OAuth. I'm instead reading the csrftoken from the website's cookie. This is the list of URLs whose cookies I can read from. The asterisks are needed for folks various ports if you want to test it locally. 

The access to read any website's title and URL is from the permissions's activeTab

rgossiaux: > @rgossiaux, didn't you say you use 0.0.0.0?

Yeah I do, I was wondering why you changed to the `*:*/*` wildcard pattern since it looked like the only missing local development host was 127.0.0.1. I think `*:*/*` could match any random site, right? I don't think we want to ask people to give us permissions to read their cookies on the whole web if we can avoid it.

maximized: @rgossiaux I thought `*:*/*` would match anything with a colon but after testing apparently not. I changed it to "http://127.0.0.1/*"  in https://github.com/cardinalitypuzzles/cardboard/pull/737

PR #737, opened 2024-01-07
Follow-up PR to allow for meta setting from Chrome Extension

maximized:


chrome_extension/cardboard/popup.js
```diff
@@ -1,10 +1,37 @@
+const IS_PROD = true;
+
+const PROD_URL = "https://cardboard.rocks";
+const DEV_URL = "http://localhost:8000";
+
+// TODO: Make this configurable
+const HUNT_NUMBER = 10;
+
+let URL = DEV_URL;
```
rgossiaux, line 9: `URL` already exists as a global, we shouldn't shadow it https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

chrome_extension/cardboard/popup.js
```diff
@@ -1,10 +1,37 @@
...
+if (IS_PROD) {
+  URL = PROD_URL;
+}
+
+// Read all puzzles and filter down to those that are metas.
```
rgossiaux, line 14: I feel like we should probably change the API to allow directly querying for metas, maybe a TODO?

chrome_extension/cardboard/popup.js
```diff
@@ -16,28 +43,40 @@ for (const tab of tabs) {
...
       mode: "cors",
       body: JSON.stringify({
         create_channels: document.getElementById("create_channels").checked,
         is_meta: document.getElementById("is_meta").checked,
         name: document.getElementById("puzzle_name").value.slice(0, 30),
         url: document.getElementById("puzzle_url").value,
+        assigned_meta: document.getElementById("puzzle_meta").value,
       }),
       headers: {
         "X-CSRFToken": cardboard_cookie.value,
```
rgossiaux, line 82: it won't let me leave a review comment on the lines directly, but you should also delete the commented section at the bottom of this file

chrome_extension/cardboard/popup.js
```diff
@@ -1,10 +1,37 @@
+const IS_PROD = true;
```
rgossiaux, line 1: Add a TODO to improve this? Feels like it should be `false` by default and then the release process for the extension includes setting it to `true` in some automatic way.

chrome_extension/cardboard/popup.js
maximized: Changed to TARGET_URL

chrome_extension/cardboard/popup.js
maximized: This would be a TODO on the API side. I added https://github.com/cardinalitypuzzles/cardboard/issues/745

chrome_extension/cardboard/popup.js
maximized: Deleted

rgossiaux (approved):
cool, sounds good

PR #743, opened 2024-01-09
add last active time

rawxfish:
Green = <5min
Yellow = 5-30min
Red = >30min

image

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -120,6 +120,7 @@ class PuzzleSerializer(serializers.ModelSerializer):
     )
     recent_editors = serializers.SerializerMethodField()
     top_editors = serializers.SerializerMethodField()
+    last_edited_on = serializers.SerializerMethodField()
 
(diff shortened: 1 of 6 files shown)


api/serializers.py
```diff
@@ -165,6 +166,15 @@ def get_top_editors(self, obj):
...
         return top_editors
 
+    def get_last_edited_on(self, obj):
+        puzzle_activities = obj.puzzle_activities.all()
+        if len(puzzle_activities) > 0:
+            return max(
+                [activity.last_edit_time for activity in obj.puzzle_activities.all()]
+            )
+        else:
+            return None
```
rgossiaux, line 176: Does `obj.puzzle_activities.order_by("-last_edit_time").first()` work? Might be better to do this in the db (not positive)

Actually I see below that you're using `prefetch_related`... hmm... in that case this seems fine

rgossiaux (approved):
nice, lgtm!

Issue #671, opened 2023-01-13
Add way to automatically create ChatRoles from default tags


rgossiaux:
Now that we have the default tags in the backend, we should add something (eg admin command) to automatically create discord roles that correspond to the default tags & to add ChatRole objects that correspond. We could probably even automate the carl-bot interaction to select the roles if we wanted.

PR #746, opened 2024-01-09
admin button to auto-gen & link Discord roles

npinsker:
image

addresses #671 

- Button on the sidebar that's visible to admins only
- Clicking it will generate Discord roles for all tags that are blue or white
- Also creates a `ChatRole` for all Discord roles, meaning people with those roles will get pinged when a puzzle's tagged with the corresponding tag


rgossiaux:
Why did you move this from the hunt edit page to the sidebar? I thought that made the most sense to me.

I also think it's unlikely we ever want to modify these during the hunt FWIW

npinsker:
moved back to `/edit`

chat/tasks.py
```diff
@@ -166,3 +166,56 @@ def handle_sheet_created(puzzle_id):
...
+            )
+            discord_roles_by_name[tag.name] = new_role_info
+            logger.info(f"Created new Discord role {tag.name}")
+
+        # Copy tag info into a ChatRole
+        existing_chat_role = existing_chat_roles.filter(name=tag.name)
+        if existing_chat_role.exists():
+            obj = existing_chat_role.first()
+            if obj.role_id != discord_roles_by_name[tag.name]["id"]:
+                with transaction.atomic():
```
rgossiaux, line 211: Does this actually do anything if we're just doing a single operation? ditto below

hunts/views.py
```diff
@@ -144,6 +144,26 @@ def redirect_to_drive(request, hunt_slug):
...
+        return HttpResponseForbidden()
+
+    hunt = Hunt.get_object_or_404(user=request.user, slug=hunt_slug)
+
+    import chat.tasks
+
+    chat.tasks.sync_roles(
+        hunt,
+        settings.CHAT_DEFAULT_SERVICE,
+    )
```
rgossiaux, line 161: I think this will run this synchronously by the web process, right? Need to call `.delay()` if you want to create a task

rgossiaux (approved):
Lgtm, nice!

PR #751, opened 2024-01-10
adds an X to the search bar and makes the ESC key clear it

### LINKED ISSUE #676: Quick way to clear search
There should be a way to clear the search filter quickly, like an X next to the search bar and/or the esc key or something

npinsker:
resolves #676


rgossiaux (approved):
Lol nice

PR #752, opened 2024-01-10
add setting for whether Discord channel creation should be on by default

### LINKED ISSUE #729: Add setting for whether Discord channel creation should be on by default
I think we wanted this off for Mystery Hunt but on for online hunts (Microsoft, Puzzle Boat)

We can add a setting to the /edit page and then pass that to the frontend via the hunt API

See eg https://github.com/cardinalitypuzzles/cardboard/commit/5fe9e5868c01cc93690081885e1e6ccddd7e5703

rawxfish:
Closes #729

`create_channel_by_default` is sorta mouthful, suggestions welcomed for a better name.


rgossiaux (approved):
yay

PR #753, opened 2024-01-10
Add per-hunt access permissions

rgossiaux:
After this change, anyone is allowed to log in to cardboard, but they can only see hunts that they have access to. For the moment, users gain access to a hunt via a button on the /edit page that syncs permissions from the Drive folder. Note that this only gives people access, it never removes access right now.


google_api_lib/tasks.py
```diff
@@ -664,3 +665,44 @@ def update_active_users(self, hunt_id):
...
+        return
+
+    hunt = Hunt.objects.select_related("settings").get(pk=hunt_id)
+
+    if not hunt.settings.google_drive_folder_id:
+        return
+
+    UserModel = get_user_model()
+    emails = get_file_user_emails.run(hunt.settings.google_drive_folder_id)
+    print(f"Found emails {emails}")
```
rgossiaux, line 705: Oops, delete these

rgossiaux:
I think I want to write some tests before trying to ship this but looking for feedback anyway

pyproject.toml
```diff
@@ -22,6 +22,7 @@ redis = "^4.1.0"
 django-celery-beat = "^2.4.0"
 python-dateutil = "^2.8.2"
 django-cors-headers = "^4.3.1"
+django-guardian = "^2.4.0"
```
rawxfish, line 25: oh cool, didn't know something like this existed

new-hunt-setup.md
```diff
@@ -131,11 +131,6 @@ Global Google Drive, Sheets, and API settings for automatic sheets creation:
 * `GOOGLE_API_PRIVATE_KEY` - the private key for the key you added, with newlines replaced with `\n` (should be the value of the `private_key` field in the downloaded JSON when you [created the key](https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating_service_account_keys); should look something like `-----BEGIN ... KEY-----\n...<long base64-encoded key>...\n-----END ... KEY-----\n`)
 * `GOOGLE_API_X509_CERT_URL` - the value of the `client_x509_cert_url` field in the downloaded JSON when you [created the key](https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating_service_account_keys)
 
-Hunt-specific Google settings:
-
-* `GOOGLE_DRIVE_HUNT_FOLDER_ID` - the id of your Google Drive folder, should be part of the URL (`https://drive.google.com/drive/folders/<folder_id>`). This environment variable is used for getting the list of allowed emails for OAuth. Please keep consistent with the value in HuntSettings. Deprecating this variable is tracked in #662.
```
rawxfish, line 136: are we able to delete the code in settings.py that read these environment variables? https://github.com/cardinalitypuzzles/cardboard/pull/753/files#diff-c003920b4e2c2658c5433b6599e1a2c0c21fd4abfeb45a7b32a8850b4a47f298L209

rawxfish: oh we still use it as the default value in [google_api_lib/tasks.py](https://github.com/cardinalitypuzzles/cardboard/pull/753/files#diff-5ea7ae2b73cb0ddaa1f9bf676e71ab5a19d755c7eea5a80e6da1df1885d7a8f0L112), probably should clean this up or make an issue to clean it up later.

rawxfish (approved):
look great, woo!

pyproject.toml
akirabaruah: I remember researching it when @rgossiaux brought it up in https://github.com/cardinalitypuzzles/cardboard/issues/191

Should we be concerned that the codebase hasn't been touched in 2+ years? https://github.com/django-guardian/django-guardian

akirabaruah: On that note, is there a reason to use `django-guardian` over the per-object permissions provided by DRF? https://www.django-rest-framework.org/api-guide/permissions/#object-level-permissions

rgossiaux: as far as I know it's the most popular/mature library for this in the django ecosystem. No commits in 2 years is probably a bit worrying but I'm just doing pretty basic core stuff with it so I'm personally not too bothered atm

DRF provides view methods for checking permissions (I'm using them in this PR already I think), but not a permissions backend itself. Still need some way of storing permissions in the DB--that's what django-guardian provides

calvinballing:
This change appears to have broken the hunt settings page by removing the submit button from the associated form

PR #755, opened 2024-01-11
Add puzzle tag selector and merge puzzle + hunt tags.

kcaze:
Addresses #672.

### DIFF
--- a/hunts/src/EditPuzzleTagsModal.js
+++ b/hunts/src/EditPuzzleTagsModal.js
@@ -13,10 +13,10 @@ import EditableTagList from "./EditableTagList";
 function EditPuzzleTagsModal({ puzzleId, puzzleName }) {
   const huntTags = useSelector(selectHuntTags);
   const puzzleTags = useSelector(selectPuzzleTags);
-  // Use a map to concat huntTags and puzzleTags and then dedupe.
+  // Use a map to concat huntTags and puzzleTags and then dedupe by id.
   const tagMap = new Map();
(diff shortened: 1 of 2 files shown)


hunts/src/EditPuzzleTagsModal.js
```diff
@@ -3,15 +3,23 @@ import { useDispatch, useSelector } from "react-redux";
...
-  const allTags = useSelector(selectHuntTags);
+  const huntTags = useSelector(selectHuntTags);
+  const puzzleTags = useSelector(selectPuzzleTags);
+  // Use a map to concat huntTags and puzzleTags and then dedupe.
+  const tagMap = new Map();
+  for (const tag of huntTags.concat(puzzleTags)) {
+    tagMap.set(tag.name, tag);
+  }
+  const allTags = Array.from(tagMap.values());
+
```
rgossiaux, line 22: I think you can dedupe by id instead of name (though since atm tags can't be renamed it's all the same, but I think id is still better)

hunts/src/puzzlesSlice.js
```diff
@@ -285,6 +285,20 @@ export const selectNumMetasUnsolved = createSelector(
   }
 );
 
+export const selectPuzzleTags = createSelector(
+  [puzzlesSelectors.selectAll],
+  (puzzles) => {
+    // Use a map to dedupe puzzle tags by name.
```
rgossiaux, line 291: ditto

rgossiaux (approved):
sweet, thank you! left a nit

PR #795, opened 2024-12-09
Simplify header puzzle stats

npinsker:
Was thinking the header displaying overall hunt progress could be cleaned up a little -- lmk what you think.

old
image

new
image


erwa (approved):
Looks good to me.

PR #799, opened 2024-12-10
reduce line height of table buttons

npinsker:
The default Bootstrap button takes up a bit more vertical space than it needs to. This reclaims a little bit of it, for puzzles without long names and lots of tags (which should be most of them).

Could be argued that this makes the page look worse (happy to abandon), but I prefer it cause it's more compact + more of a 'Google spreadsheet' look.

(some of the puzzles take up more line height than you'd expect bc of the invisible wrench button)

**old**
image

**new**
image

### DIFF
--- a/cardboard/static/style.css
+++ b/cardboard/static/style.css
@@ -103,3 +103,9 @@ body {
   max-width: 400px;
   overflow-wrap: anywhere;
 }
+
+.bootstrap button.cb-btn-compact,
+.bootstrap div.cb-btn-compact button {
+  padding: .15rem .7rem;
+  line-height: normal;
+}
--- a/hunts/src/AnswerCell.js
+++ b/hunts/src/AnswerCell.js
@@ -12,6 +12,7 @@ export default function AnswerCell({ row }) {
   if (row.original.guesses === undefined || row.original.guesses.length == 0) {
     return (
       <Button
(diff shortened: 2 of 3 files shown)


npinsker:
honestly might not matter  at all cause normally puzzles display 'last updated' status

akirabaruah (approved):
Changes look straightforward and reclaiming vertical space seems like a win. I was also thinking the buttons were looking a bit bloated. Anyway, we can always revert this if people feel strongly opposed.

akirabaruah:
> (some of the puzzles take up more line height than you'd expect bc of the invisible wrench button)

What does the invisible wrench button do? Or do you mean the "+" icon for adding multiple answers?

> honestly might not matter at all cause normally puzzles display 'last updated' status

What did you mean by the 'last updated' status here?

npinsker:
When you hover over a puzzle name, there’s a wrench button that becomes visible where you can change the name and metas. If that button spills over to another line, the vertical space will appear unnecessarily big while the button’s invisible.

Now that the buttons are a bit more compact, the column that consumes the most vertical space might be the one that says “Created 8 mins ago / Updated 3 mins ago”, so the vertical space saving in a real use case might not be as large as it is on my test instance. Not sure since I didn’t set it up.

PR #803, opened 2024-12-13
fix button compactness for dark mode

npinsker:
`.bootstrap` isn't applied in dark mode apparently


akirabaruah (approved):
Thanks for the quick fix 😎

PR #814, opened 2024-12-18
fix discord button background

rawxfish:
Also added a small border to make the disabled version of the button stand out more in light mode against the green and red backgrounds of solved and stuck puzzles.
Light mode:
image

Dark mode:
image


hunts/src/LinkCell.js
```diff
@@ -211,8 +213,8 @@ export const LinkCell = ({ row }) => {
       ) : loadingDiscord ? (
         <PulsingBlurpleDiscordSvg />
       ) : (
-        <button
-          onclick={createDiscordChannels}
+        <a
```
rgossiaux, line 216: I think I told akira to use a button instead of <a> so maybe that's why it broke. What was wrong with <button>?

rawxfish: The button inherits a bunch of Bootstrap styling that adds the ugly background color and border... that whole cell of buttons uses links instead of buttons for this reason, I guess the right thing to do is switch all of them to buttons and override the Bootstrap styling... I sometimes wonder if we should replace Bootstrap with something less opinionated...

rgossiaux: Ahhhhh ok. Ok sounds good. Yeah we should definitely just get rid of bootstrap lol

rgossiaux (approved):

PR #821, opened 2024-12-21
clear cache if update_active_users hits IntegrityErrors

rawxfish:


google_api_lib/tasks.py
```diff
@@ -620,51 +620,60 @@ def update_active_users(self, hunt_id):
...
+                PuzzleActivity.objects.bulk_update(
+                    updates, fields=["last_edit_time", "num_edits"]
+                )
 
-        hunt.last_active_users_update_time = end
-        hunt.save()
+            hunt.last_active_users_update_time = end
+            hunt.save()
+    except IntegrityError:
+        # attempt to avoid the error next time by clearing cache
```
asdfryan, line 675: Let's add a `logger.warning()` here

google_api_lib/tasks.py
```diff
@@ -620,51 +620,60 @@ def update_active_users(self, hunt_id):
...
+                        F("puzzle_id"),
+                        output_field=CharField(),
+                    )
                 )
+                .filter(user_puzzle_pair__in=user_puzzle_pairs)
+                .select_for_update()
+            )
+            old_keys = [(e.user_id, e.puzzle_id) for e in old_activities]
+
+            PuzzleActivity.objects.bulk_create(
```
asdfryan, line 642: DB `bulk_create` and `bulk_update` operations are pretty susceptible to timeout / OOM issues.

What is the scale of `latest.items()`. If it can exceed O(1000), might be worth chunking this.

google_api_lib/tasks.py
```diff
@@ -620,51 +620,60 @@ def update_active_users(self, hunt_id):
...
-            PuzzleActivity.objects.annotate(
-                user_puzzle_pair=Concat(
-                    F("user_id"), Value("-"), F("puzzle_id"), output_field=CharField()
+    try:
+        with transaction.atomic():
+            user_puzzle_pairs = [
+                f"{user_pk}-{puzzle_pk}" for (user_pk, puzzle_pk) in latest.keys()
+            ]
+            old_activities = (
+                PuzzleActivity.objects.annotate(
```
asdfryan, line 629: Do we have a db index for user/puzzle?

google_api_lib/tasks.py
rawxfish: It depends on how often the CRON job is set to run. At the intervals we set, I think its like O(10).

google_api_lib/tasks.py
rawxfish: No, should we?

rgossiaux: i feel like our tables are small enough that it's not worth it but could be wrong

google_api_lib/tasks.py
```diff
@@ -620,51 +620,61 @@ def update_active_users(self, hunt_id):
...
+                )
 
-        hunt.last_active_users_update_time = end
-        hunt.save()
+            hunt.last_active_users_update_time = end
+            hunt.save()
+    except IntegrityError:
+        # attempt to avoid the error next time by clearing cache
+        logger.warning("IntegrityError while updating active users, clearing cache")
+        cache.clear()
```
rgossiaux, line 677: I'm not like totally sold this is a good idea but if you want to merge this as a kind of bandaid it's probably fine

rgossiaux (approved):

google_api_lib/tasks.py
rawxfish: lol I agree, I think a better solution to last year's issue is to remove a puzzle's key from the cache when its deleted. I'll leave this unmerged for now and merge it in last minute if we don't do a better bandaid before Hunt.

rawxfish:
Gonna leave this unmerged since issue we ran into last year should be now fixed by #831, we can reconsider if other wonky cache issues come up.

PR #831, opened 2024-12-29
soft delete for puzzles

### LINKED ISSUE #548: Reconsider deleting chat channels when deleting a puzzle
It's nice that it cleans up after itself in the case of accidents, but on the other hand if someone makes a mistake it could be frustrating if the chat channel is dead. Maybe we should save the transcript, or only archive the chat channel if it has (non-bot) messages, or something.

rawxfish:
Makes puzzle, answers, and chatrooms models with soft deletion, so deletes are undoable. PuzzleTag is left as a normal model, because there are uniqueness constraints on them, but `PuzzleTag.objects` doesn't pick up soft deleted instances so would be sorta annoying to check every time we create a new tag that there isn't a soft deleted one already. Instead I handled deleting/restoring tags through signals.

Also adds 30 minute wait before cleaning up chats (fixes #548)


rgossiaux (approved):
awesome, looks good to me I think

PR #833, opened 2025-01-03
Add additional check to verify that chrome exists

maximized:


maximized:
I tested this locally in Firefox. Without this line, I get an error that `chrome` does not exist. After this change, the error disappears and the page can load.

rgossiaux (approved):

PR #839, opened 2025-01-10
add TypeScript + type files in `hunts/src`

npinsker:
Should be a no-op for behavior, but I did remove/cleanup a little dead-path code, parameters that were never used, etc.

Typing isn't enforced at all right now.

(removed hot reload for now, since there was a race condition with webpack and it wasn't working for me anyway)


npinsker:
I fixed the Discord display bug

hunts/src/types.tsx
```diff
@@ -0,0 +1,121 @@
...
+export interface Row<D extends object = {}> extends BaseTableRow<D> {
+  canExpand: boolean;
+  isExpanded: boolean;
+  depth: number;
+
+  getToggleRowExpandedProps: any;
+}
+
+// react-bootstrap component typing has a bug on old versions --
+// see https://github.com/react-bootstrap/react-bootstrap/issues/6819
```
rgossiaux, line 104: It looks like this is a bug specific to v2.10.3 that isn't present in v2.10.2, is this the same issue you were seeing or is there another reference somewhere?

docker-entrypoint.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/sh
 
 yarn install
-# Compiles the js bundle(s) and runs in background to hot reload
-yarn run dev &
+yarn run dev
```
rgossiaux, line 4: why remove the &? Does this not just cause this shell script to hang?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "sourceMap": true,
+    "strict": true,
+    "module": "esnext",
+    "moduleResolution": "node",
+    "target": "es5",
```
rgossiaux, line 7: No need to target es5, it's 2025 now. Think es6 is probably fine

hunts/src/types.tsx
```diff
@@ -0,0 +1,121 @@
...
+export const SafeDropdownButton = DropdownButton as any;
+export const SafeFontAwesomeIcon = FontAwesomeIcon as any;
+export const SafeForm = Form as any;
+export const SafeModal = Modal as any;
+export const SafeOverlayTrigger = OverlayTrigger as any;
+export const SafePopover = Popover as any;
+export const SafeReactTimeAgo = ReactTimeAgo as any;
+export const SafeRow = BootstrapRow as any;
+export const SafeTable = Table as any;
+export const SafeTooltip = Tooltip as any;
```
rgossiaux, line 121: These renamings make a bunch of the changes in this PR very hard to review. Can we split this up to just re-export them with the same names and then rename them separately? That way the diffs will be cleaner. 

Not sure I understand the `SafeX` naming scheme

hunts/src/types.tsx
npinsker: I was mistaken sorry, it's some sort of abstruse typing incompatibility along the same lines, but that specific bug is irrelevant

hunts/src/types.tsx
npinsker: I upgraded `react-bootstrap` and `react-time-ago` and fixed the errors, which reduces the code changes significantly.

It's possible to do those upgrades in a separate commit, but it'd be a bit painful (would have to convert all the `.tsx` files back to `.js` files) and I'd prefer to just save the work and put it all in one PR, if that works for you.

hunts/src/SubmitAnswerModal.tsx
```diff
@@ -44,14 +53,14 @@ function SubmitAnswerModal({ puzzleId, puzzleName }) {
...
-            onChange={(e) => setNewAnswer(e.target.value)}
+            onChange={(e: ChangeEvent) => setNewAnswer(e.target.value)}
           />
           <Form.Check
             type="checkbox"
             defaultChecked={false}
             label="backsolved"
-            value={backsolved}
-            onChange={(e) => {
+            value={backsolved.toString()}
```
rgossiaux, line 62: This looks really sus, is this really right?

hunts/src/TagPill.tsx
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import { Badge } from "react-bootstrap";
+
+interface TagPillProps {
+  name: string;
+  color: string;
+  selected?: boolean;
+  faded?: boolean;
+  editable?: boolean;
```
rgossiaux, line 9: I guess this should be deleted then

hunts/src/types.tsx
```diff
@@ -0,0 +1,94 @@
...
+  getToggleRowExpandedProps: any;
+}
+
+// FontAwesome typing requires the 'icon' field to be one from an
+// enumeration of predefined possible strings, which can create incompatibilities
+// between different versions -- in other words, if the `fa-solid` set of
+// icons exposes a string that a FontAwesome component doesn't know about, it'll
+// complain because the type can't be cast, even if you never use the icon.
+// Typing over a set of thousands of icons is ridiculous
+// so we just ignore it for this component
```
rgossiaux, line 93: Let's put a TODO to fix this Some Day (TM)

rgossiaux (commented):
Ok went through everything, left a few small comments but overall looks good. Bless you for doing this.

hunts/src/LinkCell.tsx
```diff
@@ -95,7 +98,6 @@ const GreenSheetsSvg = () => {
       x="0px"
       y="0px"
       viewBox="0 0 512 512"
-      style={{ enableBackground: "new 0 0 24 24" }}
```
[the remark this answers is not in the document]
npinsker: This is a deprecated property that does nothing https://udn.realityripple.com/docs/Web/SVG/Attribute/enable-background

hunts/src/SubmitAnswerModal.tsx
npinsker: image
yes unfortunately

rgossiaux (approved):
Great, thank you!

hunts/src/SubmitAnswerModal.tsx
rgossiaux: Yeah but like, shouldn't this be `checked={backsolved}` or something? Not sure what this actually does. Whatever.

npinsker:
thanks for the thorough review!

hunts/src/SubmitAnswerModal.tsx
npinsker: im sorry i have no clue (`checked` is a bool field in the component as well) but this definitely works at least...

PR #840, opened 2025-01-11
Further improve Chrome Extension functionality

maximized:
When a puzzle has already been created, you can now 
- Change its status
- Set an answer
- Assign it to a new meta
- Mark as backsolved

### DIFF
--- a/chrome_extension/cardboard/popup.html
+++ b/chrome_extension/cardboard/popup.html
@@ -7,22 +7,38 @@
     <link rel="stylesheet" href="./popup.css" />
   </head>
   <body>
-    <template id="li_template">
+    <template id="new_puzzle_template">
       <li>
-        <a>
-          <label for="puzzle_name">Puzzle Name: </label><br>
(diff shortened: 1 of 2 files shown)


maximized:
https://app.screencastify.com/v3/watch/sN5xn3xCqwjccltrBap1

This change doesn't allow you to add/remove other tags or disassociate with a meta. I think it's okay if some features are not available in the extension.

maximized:
https://drive.google.com/file/d/1LY7Mv7REh3HX_OGifWbzJCgHoKzDaxqJ/view?usp=sharing

- Added navigation to puzzle from the Google Sheet
- Added puzzle editing from the Google Sheet
- Added puzzle notes to extension

npinsker:
Agree with some features not being included in the extension -- imo the experience I'd personally shoot for is "you should have everything you need for that puzzle in one Google Sheet tab". Marking stuck/unstuck & notes is part of that (bc you may want to call for help); rearranging meta structures isn't.

npinsker (approved):
thanks for this!

PR #846, opened 2025-01-15
upgrade to bootstrap 5

npinsker:
Display issues surfaced from recently upgrading `react-bootstrap` without a corresponding upgrade to our bootstrap version (via `toggle-bootstrap`). `toggle-bootstrap` using bootstrap v5 doesn't exist, bc bootstrap v5 natively supports dark mode, so i replaced the library with vanilla.

this caused a few more CSS issues (e.g. v5 uses css custom properties), which i did my best to fix, but something may have slipped through -- sorry!

image

image

image


cardboard/templates/base.html
```diff
@@ -21,30 +17,27 @@
...
+                        document.body.setAttribute('data-bs-theme', 'light');
+                        localStorage.setItem('cardboard-theme', "light")
                     } else {
-                        // Dark
-                        document.body.classList.remove('bootstrap');
-                        document.body.classList.add('bootstrap-dark');
-                        localStorage.setItem('cardboard-theme', "bootstrap-dark")
+                        document.body.setAttribute('data-bs-theme', 'dark');
+                        localStorage.setItem('cardboard-theme', "dark")
                     }
```
rgossiaux, line 41: can simplify this now

hunts/src/DarkModeToggle.tsx
```diff
@@ -16,15 +16,11 @@ export const DarkModeToggle = () => {
...
-              localStorage.setItem("cardboard-theme", "bootstrap");
+              localStorage.setItem("cardboard-theme", "light");
+              document.body.setAttribute('data-bs-theme', 'light');
             } else {
-              // Dark
-              document.body.classList.remove("bootstrap");
-              document.body.classList.add("bootstrap-dark");
-              localStorage.setItem("cardboard-theme", "bootstrap-dark");
+              localStorage.setItem("cardboard-theme", "dark");
+              document.body.setAttribute('data-bs-theme', 'dark');
```
rgossiaux, line 23: and here

rgossiaux (approved):
Ok, think if we're gonna do this best to ship it tonight and then that gives at least a little time to find any ui bugs

PR #848, opened 2025-01-15
menu cleanup #1: tagging in tags column, metas in name column

npinsker:
move 'tag' button to tags column, which is more intuitive.

i didn't bother to add hover behavior to the tag button because i'm planning to remove these icons altogether, but am splitting the commit up a bit.

almost all of this is just c/p between different files, the work was basically already done for me

image

move meta-parenting stuff to the puzzle edit modal and remove it from the tag modal

image

image

### DIFF
--- a/hunts/src/EditPuzzleModal.tsx
+++ b/hunts/src/EditPuzzleModal.tsx
@@ -1,8 +1,10 @@
 import React, { FormEvent } from "react";
 import { Button, Form, Modal } from "react-bootstrap";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
+import { selectHuntTags } from "./huntSlice";
 import { updatePuzzle } from "./puzzlesSlice";
 import { showModal, hideModal } from "./modalSlice";
+import EditableTagList from "./EditableTagList";
 
 import type { Dispatch } from "./store";
 import type { HuntId, PuzzleId } from "./types";
@@ -29,6 +31,9 @@ function EditPuzzleModal({
   const [newIsMeta, setNewIsMeta] = React.useState(isMeta);
   const [createChannels, setCreateChannels] = React.useState(hasChannels);
   const dispatch = useDispatch<Dispatch>();
+
+  const huntTags = useSelector(selectHuntTags);
+
   const onSubmit = (e: FormEvent) => {
     e.preventDefault();
     dispatch(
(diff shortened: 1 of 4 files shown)


hunts/src/EditPuzzleModal.tsx
```diff
@@ -29,6 +31,16 @@ function EditPuzzleModal({
   const [newIsMeta, setNewIsMeta] = React.useState(isMeta);
   const [createChannels, setCreateChannels] = React.useState(hasChannels);
   const dispatch = useDispatch<Dispatch>();
+
+  const huntTags = useSelector(selectHuntTags);
```
npinsker, line 35: i just c/p this from another file but i wonder if it can be simplified

akirabaruah: I think they're actually the same type of object, but `hunt` implies "all" while `puzzle` implies individual tags.

More detail: The backend defines PuzzleTags and uses that nomenclature. Each of these PuzzleTags is associated with a single Hunt. The frontend also uses the PuzzleTag nomenclature, but also calls them `huntTags` like this: https://github.com/cardinalitypuzzles/cardboard/blob/2a84a9bac7c25e62e7333bef51ebab67ffc0e4a4/hunts/src/huntSlice.tsx#L46

We can probably just call them `tags` in the frontend instead of `puzzleTag` and `allTags` instead of `huntTags` since we don't have any other notion of tags in Cardboard AFAIK.

akirabaruah (approved):
> i didn't bother to add hover behavior to the tag button because i'm planning to remove these icons altogether, but am splitting the commit up a bit.

Yeah, with the new arrangement, changing the tag icon to a plus or something would probably make more sense visually, but if we're getting rid of the icon entirely (https://github.com/cardinalitypuzzles/cardboard/pull/851 ?), then LGTM.

hunts/src/EditPuzzleModal.tsx
rgossiaux: Yeah so this `huntTags` thing is the set of all the tags in the hunt at the time when we loaded the hunt object (which is only once at page load). The `puzzleTags` is the set of all the tags that are currently in use on a puzzle. These two sets are a bit distinct--`huntTags` will contain all the default tags even if they aren't in use, while `puzzleTags` will contain any tags that were created after the page was loaded (including meta tags).

I think combining to only a single source would mean adding some more polling just for tags which I might be a little apprehensive about although it's conceptually a lot cleaner. 

I think this code isn't going to work as a result, I think you want `puzzleTags` and not `huntTags`. I suspect that if a meta is created after the page is loaded its tag won't appear here.

rgossiaux (changes requested):

npinsker:
I verified that your bug happened, changed code to use `puzzleTags`, and verified it no longer happens

rgossiaux (approved):
Great ty. Prolly a good indicator that the underlying code could use some improvements (naming and/or commenting at the least).

PR #851, opened 2025-01-15
menu/UI cleanup #3: click-to-edit on some columns

### LINKED ISSUE #835: Simplify three different buttons & menus for 'edit puzzle', 'tag puzzle', and 'delete puzzle' into one
image

We have three separate buttons that lead to different flows for changing the name, tags, and deleting a puzzle.

image
image

It would be simpler to combine these into a single menu where you edit name, URL, tags, and have an option to delete the puzzle (with confirmation afterwards). We should simplify the three associated buttons into a single one.

npinsker:
this removes the icons entirely. instead, clickable cells will now highlight (with a yellow background) when hovered, indicating an action can be taken. (sorry idk how to take a video) (also sorry the bg is kinda ugly but it's harder to fix because of the `react-table` component we use)

image

- clicking on Name column opens up puzzle edit dialog
- clicking on Tags column opens up tag edit dialog
- clicking on the Notes column just turns the cell into an inline editable textbox instead of opening up a modal

image

resolves #835

### DIFF
--- a/hunts/src/NotesCell.tsx
+++ b/hunts/src/NotesCell.tsx
@@ -1,57 +1,75 @@
-import React from "react";
-import { showModal } from "./modalSlice";
+import React, { useState } from "react";
 import { useSelector, useDispatch } from "react-redux";
-import { faWrench, faTag } from "@fortawesome/free-solid-svg-icons";
-import ClickableIcon from "./ClickableIcon";
+import { editNotes } from "./puzzlesSlice";
 
-import huntReducer from "./huntSlice";
-
-import type { RootState } from "./store";
+import type { Dispatch, RootState } from "./store";
 import type { Hunt, Row } from "./types";
 
 export default function NotesCell({ row, value }: { row: Row; value: string }) {
   const { id: huntId } = useSelector<RootState, Hunt>((state) => state.hunt);
-  const [uiHovered, setUiHovered] = React.useState(false);
-  const dispatch = useDispatch();
+  const [uiHovered, setUiHovered] = useState(false);
+  const [ editing, setEditing ] = useState(false);
+  const [ editedNotesValue, setEditedNotesValue ] = useState(value);
+  const dispatch = useDispatch<Dispatch>();
 
   return (
     <div
-      onMouseEnter={() => {
-        setUiHovered(true);
-      }}
-      onMouseLeave={() => {
-        setUiHovered(false);
-      }}
+    onMouseEnter={() => {
+      setUiHovered(true);
+    }}
+    onMouseLeave={() => {
+      setUiHovered(false);
+    }}
+    onClick={() => {
+      if (!editing) {
+        setEditing(true);
+        setEditedNotesValue(value);
+      }
+    }}
+    style={{width: "100%", minHeight: '1.4rem', cursor: !editing ? 'pointer' : undefined, backgroundColor: uiHovered && !editing ? '#ffe579' : undefined}}
     >
-      {value}
-      <div
-        style={{
-          display: "inline-block",
-          visibility: uiHovered ? "visible" : "hidden",
-        }}
-        onMouseEnter={() => {
-          setUiHovered(true);
-        }}
-        onMouseLeave={() => {
-          setUiHovered(false);
-        }}
-      >
-        <ClickableIcon
-          icon={faWrench}
-          onClick={() =>
-            dispatch(
-              showModal({
-                type: "EDIT_NOTES",
-                props: {
-                  huntId,
+      {editing ?
+      <div style={{ display: 'flex' }}>
+        <textarea
+          autoFocus
+          value={editedNotesValue}
+          style={{ minHeight: '70px' }}
+          onChange={(e) => {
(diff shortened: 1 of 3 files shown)


hunts/src/NotesCell.tsx
```diff
@@ -1,57 +1,75 @@
...
+      <div style={{ display: 'flex' }}>
+        <textarea
+          autoFocus
+          value={editedNotesValue}
+          style={{ minHeight: '70px' }}
+          onChange={(e) => {
+            setEditedNotesValue(e.target.value);
+          }}
+          onKeyDown={(e) => {
+            if (e.key === "Enter" && (e.ctrlKey || e.shiftKey)) {
```
akirabaruah, line 41: nit: Isn't it typically Cmd+Enter for Mac users? Could we add that too?

hunts/src/NotesCell.tsx
```diff
@@ -1,57 +1,75 @@
...
-            )
-          }
+                  body: { text: editedNotesValue },
+                })
+              ).finally(() => {
+                setEditing(false);
+              });
+            } else if (e.key == "Escape") {
+              setEditedNotesValue("");
+              setEditing(false);
```
akirabaruah, line 52: I know React batches state updates, but is it at all possible for `editedNotesValue` to be reset and have `editing` remain `true` on the re-render? Would that cause the note to be wiped accidentally in a race condition?

In other words, should we `setEditing(false)` first?

hunts/src/TagCell.tsx
```diff
@@ -21,32 +22,39 @@ function TagCell({ row }: { row: Row<Puzzle> }) {
...
+          type: "EDIT_TAGS",
+          props: {
+            huntId,
+            puzzleId: row.values.id,
+            puzzleName: row.values.name,
+          },
+        })
+       );
+     }}
+     style={{
```
akirabaruah, line 44: Should we drop a `TODO` here like you did above to remember to refactor these styles into a shared CSS class?

hunts/src/NotesCell.tsx
npinsker: there's no race condition, `setEditedNotesValue("")` doesn't have any effect (since it will just be sent again when you edit the notes field next) but feels better to me stylistically / from a lifetime point of view

akirabaruah: Oh I wasn't talking about a race between exiting and editing a second time. I was thinking about this possible scenario:

```
// editedNotesValue = "...", editing = true
setEditedNotesValue("");
// editedNotesValue = "", editing = true
// re-render? -> shows blank notes
setEditing();
// editedNotesValue = "", editing = false
```

But anyway, it should be fine because:
1. It shouldn't edit the actual notes value with the blank string.
2. React state batching should update both states before re-rendering most if not all of the time.

And yeah, I agree that your original order makes more sense in terms of lifetimes.

akirabaruah (approved):

hunts/src/NotesCell.tsx
```diff
@@ -1,57 +1,76 @@
...
-
-import type { RootState } from "./store";
+import type { Dispatch, RootState } from "./store";
 import type { Hunt, Row } from "./types";
 
 export default function NotesCell({ row, value }: { row: Row; value: string }) {
   const { id: huntId } = useSelector<RootState, Hunt>((state) => state.hunt);
-  const [uiHovered, setUiHovered] = React.useState(false);
-  const dispatch = useDispatch();
+  const [uiHovered, setUiHovered] = useState(false);
```
rgossiaux, line 10: For CSS stuff I think it is much preferable to use `:hover` to control the interactions rather than to manage it in JS when possible. The JS way is kind of fragile and error prone.

hunts/src/NotesCell.tsx
```diff
@@ -1,57 +1,76 @@
...
+      <div style={{ display: 'flex' }}>
+        <textarea
+          autoFocus
+          value={editedNotesValue}
+          style={{ minHeight: '70px' }}
+          onChange={(e) => {
+            setEditedNotesValue(e.target.value);
+          }}
+          onKeyDown={(e) => {
+            if (e.key === "Enter" && (e.ctrlKey || e.shiftKey || e.metaKey)) {
```
rgossiaux, line 42: Isn't this a reverse of normal semantics? Normally enter is submit and shift+enter is new line right?

rgossiaux (changes requested):
Hmm are we confident in merging this in at this point? Think it would be good to be able to test and feel confident about this kind of change so close to hunt. @rawxfish what do you think?

Couple comments inline

hunts/src/NotesCell.tsx
npinsker: yeah i thought about this, but notes are often multiline, and i think people are going to get more confused by pressing enter and having it submit. `modifier + enter` isn't very discoverable, but there's a checkmark button at least

hunts/src/NotesCell.tsx
npinsker: sorry i jumped the gun on merging this, fixed in #854

PR #852, opened 2025-01-15
Turn off autocomplete in form modals

akirabaruah:
Prevents password managers from mistakenly autofilling names, etc.


akirabaruah:
@rgossiaux I tested with my browser autocomplete and it works. I don't use 1Password, but based on their docs, this should work for the 1Password extension too.

rgossiaux (approved):
great, thanks!

PR #857, opened 2025-01-17
Misc frontend tweaks & fixes

rgossiaux:
Fix Notes cell bugs

    The textarea was only taking up part of the width of the cell, and the
    yellow hover background would still appear while you were editing the note
    which was weird.

    Re-order Edit Puzzle modal

    This checkbox looked kinda bad underneath the meta tag section.
    Change the ugly "Is meta" language
    Hide the meta section when there are no metas to display

### DIFF
--- a/hunts/src/EditPuzzleModal.tsx
+++ b/hunts/src/EditPuzzleModal.tsx
@@ -99,7 +99,7 @@ function EditPuzzleModal({
           <h5 style={{ textAlign: "center" }}>Parent Metas</h5>
           <EditableTagList
             puzzleId={puzzleId}
-            tags={tags.filter((tag) => tag.is_meta && tag.name != name)}
+            tags={meta_tags}
           />
           </>) : undefined}
         </Modal.Body>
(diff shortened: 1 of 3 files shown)


hunts/src/EditPuzzleModal.tsx
```diff
@@ -98,6 +94,14 @@ function EditPuzzleModal({
             // If channel is already created, disable its deletion
             disabled={hasChannels}
           />
+          {meta_tags.length > 0 ? (
+              <>
+          <h5 style={{ textAlign: "center" }}>Parent Metas</h5>
+          <EditableTagList
+            puzzleId={puzzleId}
+            tags={tags.filter((tag) => tag.is_meta && tag.name != name)}
```
npinsker, line 102: `tags={meta_tags}`

npinsker (approved):
thanks!

hunts/src/EditPuzzleModal.tsx
rgossiaux: Oops thanks

PR #859, opened 2025-01-17
Make sure pre-commit runs prettier on jsx/ts/tsx files

rgossiaux:
right now it doesn't

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,5 +25,5 @@ repos:
     rev: v2.7.1
     hooks:
       - id: prettier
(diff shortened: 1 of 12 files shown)


npinsker:
well this definitely works... do you mind bundling the prettier changes into this commit too? (or do a separate commit if you prefer)

rgossiaux:
yeah I will run `pre-commit run --all-files` and attach it to this PR (need to do this to get it to pass the check anyway), I don't want to do it this very second to avoid merge conflicts with other PRs

npinsker (approved):

PR #862, opened 2025-01-17
Pin initial puzzle channel messages to Discord Channel

calvinballing:
This pins the two Discord messages with links (General Links, and Google Sheets Link) to the top of each Discord puzzle channel.  It might also be nice to pin some later messages such as when a puzzle is solved, however if a puzzle becomes unsolved that message should get unpinned, which is a little more complicated, so I skipped that for now


discord_lib/discord_chat_service.py
```diff
@@ -44,20 +44,33 @@ def _make_link_embeds(self, embedded_urls):
         # 12852794 is Cardinal
         return [{"fields": fields, "color": 12852794, "type": "rich"}]
 
-    def send_message(self, channel_id, msg, embedded_urls={}):
+    def send_message(self, channel_id, msg, embedded_urls={}, pin=False):
```
npinsker, line 47: - remember to match the [`ChatService` contract](https://github.com/cardinalitypuzzles/cardboard/blob/main/chat/service.py) to your changes

- i don't like adding that parameter to `send_message`, we should keep it separate from the ability to pin. instead, we should independently add the ability to pin a message to `ChatService`, and expose some way to get the id of the message, e.g. by just returning it from `send_message`

npinsker (changes requested):
thanks for the changes! as a heads up -- as you can imagine, we're not gonna actively merge backend PRs during Mystery Hunt

discord_lib/discord_chat_service.py
calvinballing: Your proposal would definitely be a better construction.  This was more of a quick and dirty way to get it working.  I may look at refactoring this after hunt

calvinballing:
> thanks for the changes! as a heads up -- as you can imagine, we're not gonna actively merge backend PRs during Mystery Hunt, but happy to continue reviewing and can get it in afterwards

We've been using this during our practice hunt successfully, but yeah, I wouldn't expect you to pull in untested code last minute :)

PR #863, opened 2025-01-17
Update comments

calvinballing:
Removes outdated comment


npinsker (commented):
do you mind just removing the comment entirely? i think it’s not needed

npinsker (approved):

PR #864, opened 2025-01-17
fix recreating deleted puzzles

### LINKED ISSUE #841: Handle re-creation of soft deleted puzzles
Right now we're just hitting an IntegrityError b/c uniqueness constraints. Options include restoring original puzzle or hard deleting original puzzle and re-creating, not sure which one is better.

rawxfish:
Fixes #841


npinsker (approved):
this looks fine but maybe let’s not merge yet unless necessary?

puzzles/models.py
```diff
@@ -84,7 +84,7 @@ class Puzzle(SoftDeleteModel):
     class Meta:
         constraints = [
             models.UniqueConstraint(
-                fields=["name", "hunt"], name="unique_names_per_hunt"
+                fields=["name", "hunt", "deleted_at"], name="unique_names_per_hunt"
```
akirabaruah, line 87: Should we add a comment here explaining why this is included in the uniqueness constraint?

It isn't immediately obvious how `deleted_at` is related to `unique_names_per_hunt`.

akirabaruah (approved):

PR #870, opened 2025-01-21
replace redux with zustand

npinsker:
replace redux with zustand.

i'm not an amazing programmer, but i feel i'm alright at least. but this kind of thing just makes my eyes glaze over...!!
image

i've used redux professionally and personally, and for me -- the mental model has always been too complicated. rampant functions passing around other functions and lots of random magic "effects at a distance" that seem to have the goal of trimming off single-digit lines of code... to me at least, it doesn't seem optimized for simplicity and readability. one reason galactic recently passed on adopting cardboard was because they don’t like redux.

i looked into a few replacement solutions and zustand seems like redux-but-cleaner — coupled logic is much closer together, and it doesn't have "thunks" or "reducers" (but does have slices). there's a really awesome library it recommends called "emmer" that generates immutable updates very easily and is used here to transform state. imo the biggest gains are in `puzzlesSlice`.

sorry in advance… again…

image
(i basically didn't prompt the AI for this... i just asked it if there was any reason at all to use redux... and it said no)


hunts/src/HuntViewMain.tsx
```diff
@@ -41,32 +33,79 @@ const MODAL_COMPONENTS: Record<string, any> = {
...
-  const modal = useSelector<RootState, any>((state) => state.modal);
-  const hunt = useSelector<RootState, any>((state) => state.hunt);
-  const alert = useSelector<RootState, any>((state) => state.alert);
-  const filterTags = useSelector<RootState, PuzzleTag[]>(getTagFilter);
+export interface PuzzleTable extends Puzzle {
+  subRows: Puzzle[];
+}
 
-  const filterSolved = useSelector(getSolveStateFilter);
+export const makePuzzleTableData = (ids: Puzzle[]) => {
```
npinsker: moved from `puzzlesSlice`

hunts/src/DeleteAnswerModal.tsx
```diff
@@ -28,7 +27,7 @@ function DeleteAnswerModal({
       </Modal.Header>
       <Modal.Body>Are you sure you want to delete this answer?</Modal.Body>
       <Modal.Footer>
-        <Button variant="secondary" onClick={() => dispatch(hideModal())}>
+        <Button variant="secondary" onClick={() => hideModal()}>
```
akirabaruah: Why not `onClick={hideModal}` here?

hunts/src/AddPuzzleModal.tsx
```diff
@@ -100,7 +97,7 @@ function AddPuzzleModal({ huntId }: { huntId: HuntId }) {
           />
         </Modal.Body>
         <Modal.Footer>
-          <Button variant="secondary" onClick={() => dispatch(hideModal())}>
+          <Button variant="secondary" onClick={() => hideModal()}>
```
akirabaruah: Why not `onClick={hideModal}` here?

hunts/src/DeletePuzzleModal.tsx
```diff
@@ -16,10 +13,12 @@ function DeletePuzzleModal({
...
 }) {
-  const dispatch = useDispatch<Dispatch>();
+  const { deletePuzzle } = useStore((state) => state.puzzlesSlice);
+  const { hideModal } = useStore((state) => state.modalSlice);
+
   const onDelete = () => {
-    dispatch(deletePuzzle({ huntId, id: puzzleId })).finally(() => {
-      dispatch(hideModal());
+    deletePuzzle(puzzleId).finally(() => {
+      hideModal();
```
akirabaruah: Similar comment: Why not `finally(hideModal)` here?

hunts/src/DeletePuzzleModal.tsx
```diff
@@ -32,7 +31,7 @@ function DeletePuzzleModal({
         Are you sure you want to delete <b>{puzzleName}</b>?
       </Modal.Body>
       <Modal.Footer>
-        <Button variant="secondary" onClick={() => dispatch(hideModal())}>
+        <Button variant="secondary" onClick={() => hideModal()}>
```
akirabaruah: `{hideModal}`?

hunts/src/EditAnswerModal.tsx
```diff
@@ -49,7 +41,7 @@ function EditAnswerModal({
           />
         </Modal.Body>
         <Modal.Footer>
-          <Button variant="secondary" onClick={() => dispatch(hideModal())}>
+          <Button variant="secondary" onClick={() => hideModal()}>
```
akirabaruah: `{hideModal}`?

hunts/src/EditNotesModal.tsx
```diff
@@ -46,7 +40,7 @@ function EditNotesModal({
           />
         </Modal.Body>
         <Modal.Footer>
-          <Button variant="secondary" onClick={() => dispatch(hideModal())}>
+          <Button variant="secondary" onClick={() => hideModal()}>
```
akirabaruah: `{hideModal}`?

hunts/src/filterSlice.tsx
```diff
@@ -1,53 +1,67 @@
...
+  UNSOLVED_WITH_SOLVED_METAS = 3, // Includes unsolved feeders with solved metas.
+}
+
+export interface FilterSlice {
+  filterSlice: {
+    filterValue: string;
+    solveStateFilter: SOLVE_STATE_FILTER_OPTIONS;
+    tags: PuzzleTag[];
+
+    updateFilterValue: (filterValue: string) => void;
```
akirabaruah: nit: For simplicity, why not call this `setFilterValue`? Update seems like it might have other side effects.

hunts/src/HuntViewMain.tsx
akirabaruah: Does it still need to be `export`ed?

hunts/src/filterSlice.tsx
akirabaruah: In general, I wonder if we should use `set` instead of `update` for the various setter functions in each slice.

hunts/src/utils.tsx
akirabaruah, line 1: nit: Should this be named `utils.ts`? It doesn't seem to have any React component stuff; i.e. it's just a Typescript function right?

hunts/src/filterSlice.tsx
```diff
@@ -1,53 +1,67 @@
...
+    updateFilterValue: (filterValue: string) => {
+      set((state) => {
+        state.filterSlice.filterValue = filterValue;
+      });
     },
-    updateSolveStateFilter: (state, action) => {
-      state.solveStateFilter = action.payload;
+    updateSolveStateFilter: (solveStateFilter: SOLVE_STATE_FILTER_OPTIONS) => {
+      set((state) => ({
+        filterSlice: { ...state.filterSlice, solveStateFilter },
```
akirabaruah: Why do we spread the old `filterSlice` here? Shouldn't we just set `state.solveStateFilter` to the new `solveStateFilter` similar to `updateFilterValue` above? These are both just setter functions, right?

akirabaruah (commented):
Thanks for proposing this. I agree that Zustand is simpler and nicer to use than Redux. I also wonder whether we need this much state management in the client in the first place, but this translation from Redux seems like a step in the right direction.

I noticed lots of expressions like `{() => func()}` or `{() = { func(); }}` which I think we can collapse to just `{func}`. Started notating them individually, but realized there are lots of occurrences.

Anyway, seems like a fairly straightforward translation. LGTM overall.

hunts/src/filterSlice.tsx
npinsker: you’re right, thanks for the catch — the reason is i wrote that line before i installed immer, so it was necessary then, but i forgot to update it

npinsker (commented):
responded to comments, thanks for the review!!

hunts/src/HuntViewMain.tsx
npinsker: nope

hunts/src/HuntViewMain.tsx
```diff
@@ -41,32 +33,79 @@ const MODAL_COMPONENTS: Record<string, any> = {
...
 
-  const updatePuzzleData = () => {
-    dispatch(fetchPuzzles(props.huntId));
-  };
+  const l = useStore((state) => state.puzzlesSlice.lastUpdate);
 
-  useInterval(updatePuzzleData, 10 * 1000);
+  useInterval(() => {
+    fetchAllPuzzles();
+  }, 10 * 1000);
```
akirabaruah: Oh, maybe one more of these calls that can be collapsed into `useInterval(fetchAllPuzzles, 10 * 1000);`?

hunts/src/HuntViewMain.tsx
```diff
@@ -41,32 +33,79 @@ const MODAL_COMPONENTS: Record<string, any> = {
...
+    api.getHunt(props.huntId).then((response) => {
+      hunt.set(response);
+      fetchAllPuzzles();
+    });
   }, [props.huntId]);
 
   React.useEffect(() => {
     if (alert.show) {
-      const timer = setTimeout(() => dispatch(hideAlert()), 8 * 1000);
+      const timer = setTimeout(() => hideAlert(), 8 * 1000);
```
akirabaruah: Same :)

hunts/src/HuntViewMain.tsx
```diff
@@ -120,28 +159,26 @@ export const HuntViewMain = (props: { huntId: HuntId }) => {
...
-          data={tableData}
+          data={makePuzzleTableData(Object.values(puzzles))}
           filterSolved={filterSolved}
           filterTags={filterTags}
         />
         <Modal
           animation={false}
           show={modal.show}
-          onHide={() => dispatch(hideModal())}
+          onHide={() => modal.hideModal()}
```
akirabaruah: Does `{modal.hideModal}` also work?

akirabaruah (approved):
Super clean. Thanks! Found another small cleanup below, but still LGTM overall.

We should probably ask @rgossiaux to take a look or at least rubber stamp, since he originally integrated the Redux logic IIRC.

